PR #2, opened 2020-08-10
Configure mypy and enable django stubs

jbeezley:


mgrauer:
This work should be moved to the cookie cutter.

jbeezley:
Moving upstream.

PR #9, opened 2020-08-20
Add an abs_path property to Folder and File

brianhelba:


zachmullen (dismissed):
Let's merge this, it's a highly reasonable first implementation.

We should open an issue to investigate more clever implementations.

brianhelba:
I updated this to the latest codebase, added support for File, and added tests.

It's still not used anywhere, but it seems like it might be useful at some point, and I wanted to get rid of the outstanding PR.

zachmullen (approved):

PR #41, opened 2020-10-08
Quota representation and behavior in model layer

zachmullen:
The scope of this PR is just to add the database representation of quotas, not to actually add behavior to enforce them.

I went with the existing strategy, and the one that seems to be recommended in Django land, of using a OneToOneField to effectively attach columns to user rows, but using a separate relation. The performance implications, as far as I can guess, are negligible in both directions simply due to how we use quotas. However, there are other tradeoffs that merit consideration:

**Pros:**

* More sanctioned pattern at the Django level? [This](https://simpleisbetterthancomplex.com/tutorial/2016/07/22/how-to-extend-django-user-model.html) is a good article I was reading on the subject. Not sure if it's authoritative.
* Separation of concerns in the application layer. Having the quota concept decoupled from the identity concept is definitely valuable in our Python code, though I'd still need to be convinced that it provides any value to separate it at the DBMS layer.

**Cons:**

* Violates the principle of "make invalid states unrepresentable". If what we'd really like is to ensure that every user has a used count and allowed quota associated with them (which I think we do), it would be preferable to simply add these as columns to the existing user table and make them required and provide the appropriate constraints. As written, it's possible at the DBMS layer to get into a state where a user does not have associated quota columns. We can of course enforce this consistency at the application layer, but doing so is subverting some of the value of our wonderfully powerful DBMS, which should generally make us feel less confident.
* This isn't to do with quotas specifically, but if we are never planning to extend the user model, we lose out on additional desirable capabilities, like adding extra methods or properties onto it. However, it looks like we could achieve this using a Proxy model as mentioned in the article above.
* Any operation where quotas are required will require an additional database query, or at the very least a join. Again, this is probably negligible since we don't really do bulk operations on quotas.

FWIW I'm fine with the approach here, I just want us to make an informed decision. We're taking on a little bit of debt in this approach by forcing our app layer to have more consistency checks.


dkc/core/models/user_quota.py
```diff
@@ -1,26 +1,29 @@
...
 from django_girders.models import SelectRelatedManager
 
 
-class Quota(models.Model):
-    allocation = models.BigIntegerField(default=0)
-    user = models.OneToOneField(User, null=True, on_delete=models.CASCADE)
-    # root_set
+class UserQuota(models.Model):
+    user = models.OneToOneField(
+        User, primary_key=True, on_delete=models.CASCADE, related_name='quota'
```
zachmullen, line 13: Note to reviewers, I'd like to propose this pattern as a good idea generally for one-to-one relationships like this. It removes the unnecessary surrogate key from this table and uses the target relation's pkey as a natural pkey instead.

dkc/core/apps.py
```diff
@@ -4,3 +4,5 @@
 class CoreConfig(AppConfig):
     name = 'dkc.core'
     verbose_name = 'data.kitware.com: Core'
+
+    DEFAULT_USER_QUOTA = 3 << 30  # 3 GB
```
brianhelba, line 8: Can we prefix all our settings with `DKC_`?

dkc/core/models/user_quota.py
```diff
@@ -1,26 +1,29 @@
...
+    )
+    used = models.PositiveBigIntegerField(default=0)
+    allowed = models.PositiveBigIntegerField()
 
     # The user is needed to stringify, so join it when directly querying for a quota
     objects = SelectRelatedManager('user')
 
     def __str__(self):
-        return f'Quota for <{self.user}>; {self.allocation} bytes'
```
brianhelba, line 19: I'm worried this might actually be a little long. Have you seen it rendered somewhere in the admin page yet?

dkc/core/apps.py
brianhelba: Also, I think this should simply be a setting, added in `settings.py`.

dkc/core/models/user_quota.py
```diff
@@ -1,26 +1,29 @@
...
+        return f'Quota for <{self.user}>; used {self.used} of {self.allowed} bytes'
 
 
 @receiver(post_save, sender=User)
 def create_user_quota(sender: Type[User], instance: User, created: bool, **kwargs):
     if created:
-        quota = Quota(user=instance)
-        quota.save()
+        default_quota = apps.get_app_config('core').DEFAULT_USER_QUOTA
```
brianhelba, line 28: I'm hoping this can just be an ordinary setting, rather than an extra attribute of the app config.

dkc/core/models/user_quota.py
```diff
@@ -1,26 +1,29 @@
...
-class Quota(models.Model):
-    allocation = models.BigIntegerField(default=0)
-    user = models.OneToOneField(User, null=True, on_delete=models.CASCADE)
-    # root_set
+class UserQuota(models.Model):
+    user = models.OneToOneField(
+        User, primary_key=True, on_delete=models.CASCADE, related_name='quota'
+    )
+    used = models.PositiveBigIntegerField(default=0)
+    allowed = models.PositiveBigIntegerField()
```
brianhelba, line 16: The default value can be a callable. We should define a local private function, and use it to return the setting value.

dkc/core/models/user_quota.py
```diff
@@ -1,26 +1,29 @@
...
+        return f'Quota for <{self.user}>; used {self.used} of {self.allowed} bytes'
 
 
 @receiver(post_save, sender=User)
 def create_user_quota(sender: Type[User], instance: User, created: bool, **kwargs):
     if created:
-        quota = Quota(user=instance)
-        quota.save()
+        default_quota = apps.get_app_config('core').DEFAULT_USER_QUOTA
+        UserQuota(user=instance, allowed=default_quota).save()
```
brianhelba, line 29: Behavior in a signal handler should always be a last resort. We can set the default `allowed` on the field itself, see my comment there.

dkc/core/tests/test_quotas.py
```diff
@@ -0,0 +1,9 @@
+from django.apps import apps
+import pytest
+
+
+@pytest.mark.django_db
+def test_user_creation_creates_quota(user):
+    default = apps.get_app_config('core').DEFAULT_USER_QUOTA
```
brianhelba, line 7: Note, if we make this a setting, we can use pytest-django's `setting` fixture to temporarily alter this value for testing, if desired.

brianhelba (changes requested):
Finished my initial review. See inline comments.

dkc/core/models/user_quota.py
zachmullen: It doesn't currently appear on the admin page. I may just strip some of these things out for now and let us deal with the admin page later. I'll probably also remove the SelectRelatedManager as I don't even think we'll query the quota tables directly, but rather always through FK lookups.

dkc/core/models/user_quota.py
zachmullen: Sure, this is just my ignorance of where the best place to put stuff like this is. Is there some documentation about what belongs where?

dkc/core/models/user_quota.py
brianhelba: Note, it can't be a lambda, since that can't be serialized to a migration.

zachmullen:
PTAL, I've force pushed a new initial commit intended to capture the data modeling and quota resolving behavior. The approach as written is:

1. Every user, when created, gets a quota associated with their user record. A quota is a tuple of two strictly integer values: amount used, and amount allowed.
1. Every folder has an `owner` field, meaning each and every folder in the system is considered to be owned by one user.
1. Every root folder, when created, gets a quota associated with it. A folder quota record contains a strictly integer valued amount used column, and a nullable integer allowed column. A null `allowed` column on a folder means that the allowed amount is actually the owning user's allowed amount.
1. The process of converting a folder from counting against the user's quota to instead having its own quota involves setting a non-null `allowed` amount as the folder quota.

This approach means that we will need to specially handle when an admin user flips a folder quota from null to non-null, by transferring the used amount from the owning user's pile to the folder's custom allotment. In that case it's just a matter of subtracting from the user's `used` value, as the folder's own `used` should already be stored. I haven't written this logic yet -- I plan to put all quota computation and enforcement logic on future PRs, if this approach is approved.

zachmullen:
I went ahead and stuck rudimentary quota accounting, as well as enforcement, on this branch.

dkc/core/models/folder.py
```diff
@@ -59,8 +58,56 @@ class Meta:
...
+            user_quota.save()
+            if amount > 0:
+                user_quota.refresh_from_db()
+                if user_quota.used > user_quota.allowed:
+                    raise ValidationError(
+                        'User size quota would be exceeded: '
+                        f'{user_quota.used}B > {user_quota.allowed}B.'
+                    )
+        elif amount > 0:
+            folder_quota.refresh_from_db()
```
zachmullen, line 105: Please advise on best practices here -- because I'm using an `F` expression, `refresh_from_db` needs to be called before the updated attributes will have their normally-typed values. I chose to only call `refresh_from_db` internally here if it's strictly necessary, for the sake of performance, but callers of this function would then have to be concerned with refreshing the objects themselves if they need to use these properties again (which they probably won't).

dkc/core/models/file.py
```diff
@@ -64,5 +64,13 @@ def clean(self) -> None:
 
 @receiver(models.signals.pre_save, sender=File)
 def _file_pre_save(sender, instance, *args, **kwargs):
-    # TODO this is where we might handle quotas
     instance.size = instance.blob.size
+
+    if not instance.pk:  # this is a new file; reserve quota
+        instance.folder.increment_quota(instance.size)
```
zachmullen, line 70: This is a temporary place for incrementing the quota. Ultimately we'll want to do that when they first initiate an upload, before any bytes get stored anywhere.

zachmullen:
This now has all the quota features we need for a MVP. The branch is ready for final review, however we shouldn't merge it until we solve one problem that it introduces.

This branch adds a new constraint: all folders must have an owning user. I like this constraint and want to keep it. However, since we don't currently have user authentication, the process of assigning an owner to a newly created folder doesn't exist. As such, on this branch right now, it's impossible to create folders via the REST API. Two paths forward would be:

1. Hack together something to make every folder have some arbitrary owner that we grab from the database at create time.
2. Wait to merge this branch until we have user auth in place, make the folder creation REST endpoint only open to authenticated users, and then attach it there.

dkc/core/tests/factories.py
```diff
@@ -33,6 +34,6 @@ class Meta:
     name = factory.Faker('file_name')
     description = factory.Faker('paragraph')
     blob = factory.django.FileField(data=b'fakefilebytes', filename='fake.txt')
-    creator = factory.SubFactory(UserFactory)
+    creator = factory.LazyAttribute(lambda o: o.folder.owner)
```
brianhelba, line 37:
```suggestion
    creator = factory.SelfAttribute('folder.owner')
```

Can this be simplified as [a `SelfAttribute`](https://factoryboy.readthedocs.io/en/stable/reference.html#factory.SelfAttribute)?

dkc/core/models/user_quota.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def clean(self) -> None:
+        if self.used > self.allowed:
+            raise ValidationError({'allowed': 'Must not be less than used amount.'})
+
+
+@receiver(post_save, sender=User)
+def _create_user_quota(sender: Type[User], instance: User, created: bool, **kwargs):
+    if created:
+        UserQuota(user=instance).save()
```
brianhelba, line 30:
```suggestion
        UserQuota.objects.create(user=instance)
```

This is just slightly more canonical in Django: https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.create

dkc/core/models/folder_quota.py
```diff
@@ -0,0 +1,45 @@
...
+        # TODO should we also support a folder going from having a quota to being put back
+        # toward a user quota? Never needed that case before, but might be good for consistency.
+
+        super().save(*args, **kwargs)
+
+
+@receiver(post_save, sender=Folder)
+def _create_root_folder_quota(sender: Type[Folder], instance: Folder, created: bool, **kwargs):
+    if created and instance.parent is None:
+        FolderQuota(folder=instance, allowed=None).save()
```
brianhelba, line 45:
```suggestion
        FolderQuota.object.create(folder=instance, allowed=None)
```


This is just slightly more canonical in Django: https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.create

zachmullen:
Good news... I've worked around the lack of authentication in the REST API for now. This is ready to merge!

PR #49, opened 2020-10-21
Unify Quota as a single model type

brianhelba:
@zachmullen PTAL. I think this is significant improvement in comprehensibility.

Next steps (which definitely belong as discrete features) can be:
* Recalculate quota usages, as an async task
* Adding sizes directly to every folder
* Implement a "quota ejection" feature


dkc/core/models/quota.py
```diff
@@ -0,0 +1,64 @@
...
+                check=models.Q(used__lte=models.F('allowed')),
+                name='used_lte_allowed',
+            )
+        ]
+
+    used = models.PositiveBigIntegerField(default=0)
+    allowed = models.PositiveBigIntegerField(default=_default_user_quota)
+
+    # Nullable OneToOneField ensures that only one Quota per User exists
+    user = models.OneToOneField(
```
zachmullen, line 28: I thought the intent here was that the quota relation should be decoupled from users and folders, and that users and folders could instead point to the quota table? Having a user column here violates that design.

dkc/core/models/folder.py
```diff
@@ -58,56 +65,19 @@ class Meta:
...
-                raise ValidationError(
-                    'Root folder size quota would be exceeded: '
-                    f'{folder_quota.used}B > {folder_quota.allowed}B.'
-                )
+    # Prevent deletion of User if it has Folders referencing it
+    owner = models.ForeignKey(User, on_delete=models.PROTECT)
+
+    # Prevent deletion of a Quota if it has Folders referencing it,
+    # unless Quota is also being deleted implicitly via a permitted CASCADE
+    quota = models.ForeignKey(Quota, null=True, on_delete=models.RESTRICT)
```
zachmullen, line 73: I'd prefer not to have a nullable quota field on all folders. Only a small minority of folders will have quotas associated with them. As such, I think we should add another relation that associates (root) folders with quotas.

dkc/core/models/quota.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        if amount == 0:
+            return
+
+        try:
+            # Use an .update query instead of a .save, to avoid assigning an F-expression on the
+            # local instance, which might need to be rolled back on a failure
+            Quota.objects.filter(pk=self.pk).update(used=(models.F('used') + amount))
+        except IntegrityError as e:
+            if '"used_lte_allowed"' in str(e):
```
zachmullen, line 50: 👍

zachmullen, line 50: 👍  for switching to catching the constraint error... is there not a better way than substring matching to test what constraint failed?

dkc/core/models/quota.py
brianhelba: I sort of agree in principle. The practical difficulty is that using a custom Django `User` model is quite burdensome. The better approach is to create a `Profile` model, point it at a `User` with a `OneToOneField` relation, then tack your application-specific fields onto the `Profile`. However, even adding a new `Profile` model adds some amount of engineering burden (particularly around its automatic creation, usage with test factories, etc.), so I took this simpler approach for now. I'm sure we'll eventually need a `Profile`, but it'd be easier to add incrementally.

One ostensible weakness here is that we don't have a database-level guarantee that every `User` has a `Quota`. However, even with a `Profile,` we actually don't have that guarantee either (we can guarantee that every `Profile` has a `Quota`, but not that every `User` has a `Profile` and hence a `Quota`).

One surprising advantage of this method is that it's very easy to filter `Quota` by which have a null `user`. This allows us to distinguish "project quotas" from "user quotas", which might be useful administratively, and which is not practical if the relation goes in the reverse direction. Once I saw is, I actually realize that semantically, it does kind of make sense for `Quota` to explicitly "know about" `User`, since we often talk about "user quotas" in our design discussions.

dkc/core/models/folder.py
brianhelba: I'm not sure of an elegant way to engineer this in Django's ORM which gives us the benefits of this behavior. Right now, we can seamlessly say `isinstance(root_folder.quota, Quota)` `child_folder.quota is None`. Additionally, Django certainly knows about nullable `ForeignKey` and can do predictable things when joining them, caching them, or doing a `select_related`.

A `ManyToManyField` does create an intermediate table (relation), but its engineering does not anticipate that one column might practically be unique. We'd have to add some extra constraints to ensure this, and the default API would return [a `RelatedManager`](https://docs.djangoproject.com/en/3.1/ref/models/relations/#django.db.models.fields.related.RelatedManager) on `folder.quota`. Of course, we could add even more code to wrap this under abstraction, but that would introduce more complexity, require more testing, and we might loose some hidden optimizations from Django.

The primary cost to the current design seems to be the width of a frequently-useless integer column. Note that we are still able to provide a database-level check constraint to ensure its integrity (it is nullable XOR with `parent`). Is there's another major downside to the current approach to be considered?

dkc/core/models/quota.py
brianhelba: I'm not aware of one, since this is coming from deep within the database layer. Examining the exception, it doesn't have any properties other than `args[0]`, which looks like:
```
'new row for relation "core_quota" violates check constraint "used_lte_allowed"\nDETAIL:  Failing row contains (73, null, 3221225472, 3221225473).\n'
```

dkc/core/models/quota.py
zachmullen: > One surprising advantage of this method is that it's very easy to filter Quota by which have a null user. This allows us to distinguish "project quotas" from "user quotas"

Such semantics for null value strikes me as poor database design -- some would even call it an anti-pattern in RDBMS. The better way to distinguish project quotas from user quotas would be to do so explicitly, rather than trying to force it using a null value. I realize this stems from an idiosyncrasy of Django, but I think it would be unwise for us to let such an idiosyncrasy drive our database schema in a direction we don't really want it to go.

dkc/core/models/folder.py
zachmullen: I appreciate the value of being able to enforce the root-folder-only property of the quota using a check constraint. I'm just not convinced that it outweighs the cons associated with coupling these concerns together and making a messier schema. In practice, use of the Django ORM causes `select *` queries to be issued, so adding columns that we rarely care about is adding a little bit of cost for everyone on every fetch.

Ultimately though, I'm not trying to make a performance argument here. I just think we should have some default best practices when it comes to database normalization, and one of those practices is to avoid abuse of NULLs, and to try and separate concerns across relations. I like check constraints a lot, though -- and to be clear, in SQL land, we could write SQL functions that can execute arbitrary, multi-table queries and use them as check constraints, but Django doesn't currently have first-class support for that.

dkc/core/models/quota.py
zachmullen: I couldn't find any way either... `e.__cause__` is a `psycopg2.errors.CheckViolation`, which is more specific, but unfortunately that also doesn't provide any straightforward way of picking out which check constraint failed.

dkc/core/models/folder.py
jbeezley: > I like check constraints a lot, though -- and to be clear, in SQL land, we could write SQL functions that can execute arbitrary, multi-table queries and use them as check constraints, but Django doesn't currently have first-class support for that.

No, check constraints are expected to be immutable functions, so you can't perform subqueries in them.  https://www.postgresql.org/docs/12/ddl-constraints.html#DDL-CONSTRAINTS-CHECK-CONSTRAINTS

zachmullen: Ah, lame. I guess that only works for triggers.

dkc/core/models/file.py
```diff
@@ -67,10 +65,10 @@ def _file_pre_save(sender, instance, *args, **kwargs):
     instance.size = instance.blob.size
 
     if not instance.pk:  # this is a new file; reserve quota
-        instance.folder.increment_quota(instance.size)
+        instance.folder.effective_quota.increment(instance.size)
```
zachmullen, line 68: I think until we see an implementation of the behavior of transferring a folder from a user quota to its own quota, this isn't an apples-to-apples comparison with my approach. I think that operation is likely to be one area where your proposed approach is less elegant.

dkc/core/models/folder.py
brianhelba: > In practice, use of the Django ORM causes select * queries to be issued, so adding columns that we rarely care about is adding a little bit of cost for everyone on every fetch.

Just FYI generally, you **can** reassign the default `objects` Manager to run [`.defer`](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#defer) by default on every query.

PR #50, opened 2020-10-22
Improve the Quota admin page

brianhelba:
* Allow sorting by % used
* Humanize byte sizes in listing


brianhelba:
Putting this in a separate branch for now, to simplify review of #49 .

PR #51, opened 2020-10-22
Clean up the parameters of signal receivers

brianhelba:


dkc/core/models/file.py
```diff
@@ -63,6 +61,6 @@ def clean(self) -> None:
 
 
 @receiver(models.signals.pre_save, sender=File)
-def _file_pre_save(sender, instance, *args, **kwargs):
+def _folder_pre_save(sender: Type[File], instance: File, **kwargs):
```
zachmullen, line 64: @brianhelba was this function name change intentional?

brianhelba: No, got stuck in an ugly rebase. I'll fix.

Thanks for the catch.

PR #54, opened 2020-10-23
Use recursive CTE for path_to_root

zachmullen:
@jbeezley @brianhelba I was planning to implement folder size recording via recursive CTE, but before doing so I wanted to first test the water of doing a recursive CTE in Django so we can 1) decide if they're generally acceptable for these kinds of tree operations, and 2) establish some best practices and styles for how to use them in a safe and maintainable way. PTAL.

### DIFF
--- a/dkc/core/models/folder.py
+++ b/dkc/core/models/folder.py
@@ -71,6 +71,7 @@ class Folder(TimeStampedModel, models.Model):
         Returns a RawQuerySet that provides the path up the tree, starting with this folder
         and going all the way to the root.
         """
+        # Using LIMIT ensures this will terminate if a cycle erroneously exists
         return Folder.objects.raw(
             'WITH RECURSIVE ancestors AS ('
             'SELECT * FROM core_folder WHERE id=%s'
--- a/dkc/core/models/folder.py
+++ b/dkc/core/models/folder.py
@@ -77,8 +77,8 @@ class Folder(TimeStampedModel, models.Model):
             ' UNION ALL'
             ' SELECT f.* FROM core_folder f JOIN ancestors a ON f.id=a.parent_id'
             ')'
-            ' SELECT * FROM ancestors',
-            [self.pk],
+            ' SELECT * FROM ancestors LIMIT %s',
+            [self.pk, Folder.MAX_TREE_HEIGHT + 1],
         )
 
     def clean(self) -> None:
--- a/dkc/core/models/folder.py
+++ b/dkc/core/models/folder.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Iterator, Type
+from typing import Type
 
 from django.core import validators
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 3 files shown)


dkc/core/models/folder.py
```diff
@@ -62,16 +62,17 @@ class Meta:
...
-        return path[::-1]
+        return Folder.objects.raw(
+            'WITH RECURSIVE ancestors AS ('
+            'SELECT * FROM core_folder WHERE id=%s'
+            ' UNION ALL'
+            ' SELECT f.* FROM core_folder f INNER JOIN ancestors a ON f.id=a.parent_id'
+            ')'
+            ' SELECT * FROM ancestors',
+            [self.pk],
+        )[::-1]
```
brianhelba, line 75: This is easy to fix, but I think native order of iteration (root last) actually makes more sense for most usages.

brianhelba:
I'm not the best person to review the SQL itself. Perhaps @jbeezley or @danlamanna can take a look.

brianhelba:
@zachmullen Would you be ok to merge #56 first, which has a more useful API (which I need for a few other upcoming PRs) and a test.

If we like this PR, it should be able to somewhat seamlessly replace the implementation in #56.

dkc/core/models/folder.py
```diff
@@ -62,16 +62,17 @@ class Meta:
     # # Prevent deletion of quotas while a folder references them
     # quota = models.ForeignKey(Quota, on_delete=models.PROTECT)
 
+    @property
```
brianhelba, line 65: I'm not sure about what a good practice on properties vs traditional methods should be. My personal heuristic is:
1. Requires parameters -> method
2. Has any side effects -> method
3. Is very cheap (i.e. no database query) -> property
3. Is very expensive -> method
4. Otherwise -> ???

dkc/core/models/folder.py
```diff
@@ -62,16 +62,17 @@ class Meta:
...
     def path_to_root(self) -> List[Folder]:
```
brianhelba, line 66: Assuming the iteration reversal is dropped, this returns a `RawQuerySet[Folder]`, which is actually ideal, since it's slightly more lazy if we need it to be.

dkc/core/models/folder.py
brianhelba: Now that I think more, I agree that this particular case should be a property. It's very similar to the `RelatedManager` that tack `QuerySet`-ish properties onto models.

dkc/core/models/folder.py
zachmullen: Good point, especially given the name of the method.

zachmullen:
@jbeezley @brianhelba I've rebased and addressed feedback, this is ready for final review.

dkc/core/models/folder.py
```diff
@@ -63,12 +63,23 @@ class Meta:
...
         return self.parent is None
 
-    def iter_ancestors(self) -> Iterator[Folder]:
-        """Iterate up to the root of the Folder hierarchy, starting with this Folder."""
-        folder = self
-        while folder is not None:
-            yield folder
-            folder = folder.parent
+    @property
+    def ancestors(self) -> models.query.RawQuerySet:
```
brianhelba, line 67:
```suggestion
    def ancestors(self) -> models.query.RawQuerySet[Folder]:
```

Can you see if this works? Sometimes Python is unhappy with these template types, but this is more accurate.

brianhelba (approved):
:100:

dkc/core/models/folder.py
```diff
@@ -63,12 +63,23 @@ class Meta:
...
+        """
+        Get the path from this folder to the root folder.
+
+        Returns a RawQuerySet that provides the path up the tree, starting with this folder
+        and going all the way to the root.
+        """
+        return Folder.objects.raw(
```
brianhelba, line 74:
```suggestion
        # Using LIMIT ensures this will terminate if a cycle erroneously exists
        return Folder.objects.raw(
```

PR #55, opened 2020-10-23
Create a Tree model, holding data on Folders with a common root

brianhelba:
For now, this is an RFC, to see if this will be a useful foundation. Several minor issues need to be addressed before merging; see inline comments.


dkc/core/migrations/0010_auto_20201023_0128.py
```diff
@@ -0,0 +1,35 @@
...
+        ),
+        migrations.AlterField(
+            model_name='file',
+            name='creator',
+            field=models.ForeignKey(editable=False, null=True, on_delete=django.db.models.deletion.SET_NULL, to=settings.AUTH_USER_MODEL),
+        ),
+        migrations.AddField(
+            model_name='folder',
+            name='collection',
+            field=models.ForeignKey(default=None, on_delete=django.db.models.deletion.CASCADE, related_name='all_folders', to='core.collection'),
```
brianhelba, line 32: Note that this default is actually invalid. We can address in 2 ways:
* Squash all the migrations, and require that existing databases be dropped.
* Create a 3-step migration:
  * Add the field, but make it nullable
  * `RunPython`, where we can traverse in a more controlled ordering and populate the field
  * Alter the field, making it non-nullable; the default specified here should never actually be used

dkc/core/models/collection.py
```diff
@@ -0,0 +1,17 @@
...
+from typing import TYPE_CHECKING
+
+from django.db import models
+
+if TYPE_CHECKING:
+    # Prevent circular import
+    from .folder import Folder
+
+
+class Collection(models.Model):
```
brianhelba, line 12: I'm open to alternative names. As discussed in chat:
* I object to `RootFolder`, since this is **not** a folder, and it's confusing to reference in code
* If we want to call this `Root`, then we need to invent a new adjective for describing "folders with no parent", and rename the `Folder.is_root` property
* Other names, like `Hierarchy`, `Tree`, or anything that describes "a set of folders with a common root ancestry" would be fine

dkc/core/tests/test_metadata.py
```diff
@@ -22,6 +22,7 @@ def test_invalid_metadata(metadata, folder_factory):
     assert err.value.error_dict['user_metadata'][0].message == 'Must be a JSON Object.'
 
 
+@pytest.mark.xfail
```
brianhelba, line 25: This failure needs to be addressed, probably by making this a database test, and manually saving and adding a `Collection` to the `Folder` before attempting the `full_clean`.

Other tests that call `full_clean` are actually encountering this error too, though they are always calling `full_clean` within a `ValidationError` capturing context, so the error goes unnoticed. To be as strict as possible in what errors we are trying to trigger in those test cases, we should perhaps add similar workarounds theres.

dkc/core/migrations/0010_auto_20201023_0128.py
zachmullen: I'm a fan of dropping existing dev databases. I've already ended up doing it several times.

dkc/core/models/collection.py
zachmullen: I actually had a different vision for this model:

```python
class RootFolder(models.Model):
    folder = models.ForeignKey(Folder, primary_key=True, related_name='root_folder')
```

In that case it is precisely an enumeration of the folders that are root folders. From the doc I linked you yesterday:

> * Its pkey would be the folder ID foreign key, with cascade deletion. This is how we can ensure that root-folder-attached entities are only associated with root folders. Then the only problem we need to solve is making sure that the root folder rows only correspond to root folders, which at the moment is trivial given that folder parentage is immutable
> * For things that are required to be associated with every root folder, e.g. quotas, we can add non-nullable FKs in this root folder table
> * For things that are optionally associated with some root folders, e.g. terms, we can have a non-nullable FK from them pointing to the root folder table

On folder post-save, if `instance.is_root`, we create a record in this new relation. Does that make sense?

dkc/core/migrations/0010_auto_20201023_0128.py
brianhelba: Will do.

dkc/core/models/collection.py
brianhelba: I think the advantage of this approach is that every Folder has a direct, Django-friendly ability to reference the Tree object, and it doesn't add any extra columns (beyond what we had before). I imagine that quite often, we'll want to do `some_folder.tree.quota` or the like. This is significantly more direct than `some_folder.root_folder.tree.quota`.

Semantically, all the child Folders are members of the common Tree, and governed by the particular attributes (quota, access control, terms of use) of that Tree instance, so it's sensible to associate them directly.

zachmullen: I was thinking that `Folder.root_folder` would also be a FK into this new table. That complicates things a bit, but is good for integrity.

brianhelba: Under the approach in this PR, the strict notion of a "root Folder" is no longer meaningful. Folders can be part of Trees, which mean they indeed have a common ancestor / root, by policy, and a common set of properties.

We still get a constraint guaranteeing that every Folder references a Tree. Admittedly, the one gap in integrity constraints comes at deletion-time: we can't guarantee in the database that deleting all the Folders will delete the associated Tree; that behavior is just enforced in Python. Deleting a Tree does cascade to deleting all the Folders though, and operationally, I'd expect this might pan out to be a better way to destroy whole Trees anyway.

We still get a query to go from a Tree to a root Folder, it's just not strictly a foreign key lookup.

---

I understand the approach you were proposing, and I started implementing it that way, but I saw some downsides.

Naively, it's an additional hop to get to the Tree object, which we'll frequently want to do. "`folder` -> `root_folder` -> `tree`" vs "`folder` -> `tree`"

Of course, if the PK of the Tree table is defined to be the same as the `root_folder`, we can in theory take not actually fetch the `root_folder` row, and instead fetch the `tree` object directly. However, what then does the actual constraint on `root_folder` enforce? Does it point to a `Folder` row or a `Tree` row? We might be able to overload a single column with multiple foreign key constraints; [PostgreSQL's docs](https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-FK) don't seem to be clear on this, this StackOverflow post](https://stackoverflow.com/a/42285952) says you can do it, but we'd still need to implement our own query for one of the two references in Django, and we wouldn't be able to take advantage of Django's automatic foreign key resolution features for both. This seems like a messy road to go down practically.

Finally, how often do we care about having a direct reference from an arbitrary Folder back to the root Folder of the Tree? Now that we have `.ancestors` we *could* get there in a single (albeit less efficient) query, but the root Folder no longer has any policies or behaviors attached to it.

brianhelba:
@zachmullen Code-wise, this is complete and ready for review. I'm happy to continue discussing the fundamental architecture in addition to the implementation.

dkc/core/models/collection.py
zachmullen: > However, what then does the actual constraint on root_folder enforce? Does it point to a Folder row or a Tree row? 

I'm proposing it would point to a `Tree` (what I called `RootFolder`) row. The key's value itself is actually the same under my proposal, because I was suggesting that the Tree/RootFolder model have `folder = models.ForeignKey('Folder', primary_key=True)`. So the lookup isn't `folder -> root_folder -> tree`, it's `folder -> root_folder -> folder`. I was imagining the sorts of helpers we could use to make that less awkward and more performant, and tried a few things, but didn't come up with many great ideas.

What I want from this table is a definitive list of which folders are root folders, so that things that should only point to root folders, can point to this special relation instead. Then, rather than each root-folder-attached behavior having to do its own consistency enforcement to make sure it can only be attached to a root folder, we'll get a lot of that for free, and can manage consistency in just one place in python, i.e. when a folder's parent is being transitioned between null and non-null values.

zachmullen: Also I agree that we'll rarely care about actually getting a reference to the root folder if we attach policies to this new table -- it's not the folder itself we care about but the attached objects. In that sense, I'm actually good with the name `Tree`. I think it merits a docstring explaining its name and purpose.

But I still think it needs to be a FK back to Folder.

brianhelba: > What I want from this table is a definitive list of which folders are root folders

This is already a simple query: `Folder.objects.filter(parent=None)`

---

> I'm proposing it would point to a Tree

> But I still think it needs to be a FK back to Folder.

I don't think satisfying both of these is technically possible, unless one of the two relationships is nullable.

---

> What I want from this table is a definitive list of which folders are root folders, so that things that should only point to root folders, can point to this special relation instead

> Also I agree that we'll rarely care about actually getting a reference to the root folder if we attach policies to this new table

I'm not clear on how to reconcile these two statements conceptually.

zachmullen: > I don't think satisfying both of these is technically possible, unless one of the two relationships is nullable.

It's possible with either deferrable foreign keys in a transaction, or CTEs. (We'll need to be creating folders inside a transaction anyway since there's a cascade of related objects that will also be created in the case of root folders.) 

But it's admittedly kind of a pain to use these great features of SQL in Django, because we'd have to manually interrogate the sequence nextval.

> I'm not clear on how to reconcile these two statements conceptually.

I'll just say that for consistency, having the Tree's primary key be a pointer to its root folder is very nice. BUT... it's probably not worth the engineering effort at this time. I'm comfortable managing this complexity as written.

The only thing I'd like to consider before we merge this branch, then, is what's the best mechanism to ensure that the creation of a new root folder, along with all its hooked-in behaviors, always happens inside a txn?

zachmullen:
@brianhelba I added a commit onto the head of your branch that shows a working implementation of what I was talking about. We can roll it back, but I wanted to show it in code.

brianhelba:
@zachmullen Much appreciated, thanks for putting the idea into code. Let's discuss.

dkc/core/models/collection.py
brianhelba: > The only thing I'd like to consider before we merge this branch, then, is what's the best mechanism to ensure that the creation of a new root folder, along with all its hooked-in behaviors, always happens inside a txn?

Since Folder holds a non-nullable FK reference to Tree, a valid Tree must always exist and be pointed to before any Folder is created; I was doing this on `pre_save`. In the specific case of a root Folder, it will have to create its own Tree in `pre_save`, so integrity of having access to a Tree (and all its hooked-in behaviors) is guaranteed.

---

There are two cases that are not fully covered by my implementation:
1. When creating a root Folder, the actual initial Folder save operation fails (which could happen for lots of reasons, like a duplicate name) and no `Folder` is created, an empty Tree could be created and not garbage collected.
2. When deleting the last Folder of a Tree (not doing so by deleting the Tree itself), a Python code bug could cause the Tree to not be garbage collected.

I'm not very concerned about 2., simply because it only occurs if our code fails, and we'll know about the failure due to Sentry.

However, 1. could probably be caused by bad user input and is likely to occur at some point. The effects are ultimately harmless, since an unreferenced Tree doesn't have any effects. One option is to simply create a periodic admin task to garbage collect empty Trees if we care about the space usage. Another is to look into hooks to detect failure of the initial save and explicitly destroy the new Tree. Finally, @zachmullen 's proposed solution would ensure that a lonely Tree could not exist without any Folders referencing it, but it comes with a significant code / maintainability burden.

zachmullen: I consider periodic admin tasks to be a bigger maintainability burden than what I've written. But that's kind of moot -- when we create a new root folder, we need to create multiple documents. If the creation of one fails, the creation of others should also not succeed. This is the textbook case for using a transaction, and it's extremely easy and straightforward to do so with Django's decorators. We just have to put the transaction wrapping in the correct place to make our own lives easier. Regardless of whether we do my approach or yours for the Tree primary key, we *should* ensure that either both rows get created, or neither one. Periodic tasks that go through and clean up database integrity are kind of an admission of failure, IMO. We have ACID, so let's make use of it.

brianhelba: Agreed, being rigorous to always do (root) Folder creation in a transaction solves the problem.

dkc/core/models/folder.py
```diff
@@ -101,6 +102,21 @@ def increment_size(self, amount: int) -> None:
...
+                cursor.execute("SELECT nextval('core_folder_id_seq')")
+                self.pk = cursor.fetchone()[0]
+
+            kwargs['force_insert'] = True
+
+            with transaction.atomic():
+                self.tree = Tree.objects.create(root_folder=self)
+                super().save(*args, **kwargs)
+        else:
+            super().save(*args, **kwargs)
```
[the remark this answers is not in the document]
zachmullen: Sounds good. When you say dedicated factory method, do you mean like adding a `create_root_folder` method into this model class?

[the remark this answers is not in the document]
jbeezley: Yes, something like that.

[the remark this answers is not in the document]
zachmullen: I'll rewrite this commit to that effect. Thanks.

dkc/core/models/folder.py
```diff
@@ -57,9 +59,8 @@ class Meta:
     parent = models.ForeignKey(
         'self', on_delete=models.CASCADE, blank=True, null=True, related_name='child_folders'
     )
-    root_folder = models.ForeignKey(
-        'self', on_delete=models.DO_NOTHING, blank=True, null=True, related_name='+'
-    )
+
+    tree = models.ForeignKey(Tree, on_delete=models.CASCADE, related_name='all_folders')
```
zachmullen, line 63: Please add `editable=False`

zachmullen (changes requested):

dkc/core/models/folder.py
brianhelba: Done.

zachmullen (approved):

zachmullen (approved):

PR #57, opened 2020-10-23
Store size on folders

brianhelba:


dkc/core/models/folder.py
```diff
@@ -82,6 +84,25 @@ def ancestors(self) -> models.query.RawQuerySet[Folder]:
...
+        If this increment would exceed the max allotment for this quota, a ``ValidationError``
+        is raised and the transaction is rolled back.
+        """
+        ancestor_pks = [folder.pk for folder in self.ancestors]
+
+        Folder.objects.filter(pk__in=ancestor_pks).update(used=(models.F('size') + amount))
+
+        # Update with the new value
+        # TODO: Try without this, test if the parent is cached
+        # self.refresh_from_db(fields=['used'])
```
zachmullen, line 104: I think we'll need this, right?

dkc/core/models/folder.py
```diff
@@ -82,6 +84,25 @@ def ancestors(self) -> models.query.RawQuerySet[Folder]:
             [self.pk, Folder.MAX_TREE_HEIGHT + 1],
         )
 
+    @transaction.atomic
```
zachmullen, line 87: We should consider whether a read-committed txn isolation level is sufficient to avoid race conditions here. I know we don't yet support folder move, but with the understanding that we might someday, I think it's possible that there's a race condition if a folder was moved while one of its descendants was being uploaded into. I believe a repeatable-read isolation level would fix that.

zachmullen: After thinking about this more, I actually think read committed may be sufficient. At least, I'm fine with merging this as-is, until we see what a move implementation would look like in terms of queries.

dkc/core/models/file.py
```diff
@@ -62,5 +62,15 @@ def clean(self) -> None:
 
 @receiver(models.signals.pre_save, sender=File)
 def _file_pre_save(sender: Type[File], instance: File, **kwargs):
-    # TODO this is where we might handle quotas
-    instance.size = instance.blob.size
+    # TODO if we allow changing a file's blob, we also need to update the quota
```
zachmullen, line 65:
```suggestion
    # TODO if we allow changing a file's blob, we also need to update the size
```

dkc/core/models/file.py
```diff
@@ -62,5 +62,15 @@ def clean(self) -> None:
...
+
+    if not instance.pk:
```
zachmullen, line 67: The third argument passed to this hook is `created: bool`, which we should use instead of this check.

dkc/core/models/folder.py
```diff
@@ -82,6 +84,25 @@ def ancestors(self) -> models.query.RawQuerySet[Folder]:
...
+        """
+        Increment or decrement the Folder size.
+
+        This function increments (or decrements, if ``amount`` is negative) the Folder's current
+        size.
+
+        If this increment would exceed the max allotment for this quota, a ``ValidationError``
+        is raised and the transaction is rolled back.
+        """
+        ancestor_pks = [folder.pk for folder in self.ancestors]
```
zachmullen, line 98:
```suggestion
        if amount == 0:
            return

        ancestor_pks = [folder.pk for folder in self.ancestors]
```

dkc/core/models/folder.py
zachmullen: Or better yet, can we somehow invalidate the whole folder object without hitting the database, such that it'll be lazily refreshed if anyone tries to access attributes of it?

zachmullen (changes requested):

dkc/core/models/folder.py
```diff
@@ -45,6 +45,8 @@ class Meta:
         editable=False,
     )
 
+    size = models.PositiveBigIntegerField(default=0)
```
zachmullen, line 48:
```suggestion
    size = models.PositiveBigIntegerField(default=0, editable=False)
```

zachmullen (changes requested):

dkc/core/models/folder.py
brianhelba: @zachmullen I had the same concerns, and reached the same conclusion about waiting.

dkc/core/models/file.py
brianhelba: `pre_save` doesn't have that argument: https://docs.djangoproject.com/en/3.1/ref/signals/#pre-save

Only `post_save` has it.

zachmullen: Oops, forgot this was pre_save.

dkc/core/models/folder.py
```diff
@@ -82,6 +84,28 @@ def ancestors(self) -> models.query.RawQuerySet[Folder]:
...
+    @transaction.atomic
+    def increment_size(self, amount: int) -> None:
+        """
+        Increment or decrement the Folder size.
+
+        This function increments (or decrements, if ``amount`` is negative) the Folder's current
+        size.
+
+        If this increment would exceed the max allotment for this quota, a ``ValidationError``
+        is raised and the transaction is rolled back.
```
zachmullen, line 96: Was this comment copypasted from the quota branch? This last sentence should be removed, right?

Also this comment should mention that this increments the folder and all its ancestors.

zachmullen (changes requested):
One change needed in the docstring, but otherwise this LGTM. Just needs the migration, as well as some good tests.

dkc/core/models/folder.py
brianhelba: In the case of a simple field (like `size` here), I *believe* doing `del folder.size` is sufficient to make the field deferred, such that it will be feteched lazily on future access, the same as if it had been fetched initially [using `.defer`](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#defer).

I'm not able to find a clean, easy way to invalidate / re-defer foreign key references after they've been resolved.

However, as far as I can find, deferring fields directly on a model like this after instantiation is not documented by Django and not officially supported. In a case where we really need it for performance, I'd be open to the idea, but for now, I think we'd be deviating from standard Django usage in the name of premature optimization.

brianhelba:
I've updated this with fixes and tests. @zachmullen PTAL.

dkc/core/models/folder.py
```diff
@@ -82,6 +84,23 @@ def ancestors(self) -> models.query.RawQuerySet[Folder]:
             [self.pk, Folder.MAX_TREE_HEIGHT + 1],
         )
 
+    def increment_size(self, amount: int) -> None:
```
zachmullen, line 87: You removed the transaction decorator from this method, was that intentional?

brianhelba: Yeah, I was about to comment on that too. I think it's fine as-is, but I'd like some critical analysts.

The update itself is necessarily going to happen in it's own transaction, since it's a single statement. Until we have folder moves, the list of ancestor PKs will never change. Any subfolders added concurrently will be initialized with 0 size regardless.

It's possible now that between the update and the model reload, the size changes further in the database due to concurrent operations, but is that a problem? The new size would not be wrong, it's just more up-to-date than we'd expect. I don't think that's a problem for later local inconsistencies, since any other local models are outdated anyway.

zachmullen: > The update itself is necessarily going to happen in it's own transaction, since it's a single statement.

That's not true, it can be a part of a larger transaction. In fact, the behavior we probably want here is `SELECT FOR UPDATE` inside of a transaction block, then the update query. Django has `select_for_update`, but I'm not sure it could be applied to the RawQuerySet returned from `self.ancestors` (I would assume it cannot.)

> It's possible now that between the update and the model reload, the size changes further in the database due to concurrent operations, but is that a problem? The new size would not be wrong, it's just more up-to-date than we'd expect. I don't think that's a problem for later local inconsistencies, since any other local models are outdated anyway.

No, this isn't a problem.

zachmullen: One thing we could do is just add `FOR UPDATE` into our `ancestors` query. It would only hold onto locks while part of a transaction, and that would probably only occur if someone was making an update to the ancestors.

brianhelba: For what it's worth, [from the Django docs](https://docs.djangoproject.com/en/3.1/topics/db/transactions/#why-django-uses-autocommit):
> In the SQL standards, each SQL query starts a transaction, unless one is already active. Such transactions must then be explicitly committed or rolled back.
> 
> This isn’t always convenient for application developers. To alleviate this problem, most databases provide an autocommit mode. When autocommit is turned on and no transaction is active, each SQL query gets wrapped in its own transaction. In other words, not only does each such query start a transaction, but the transaction also gets automatically committed or rolled back, depending on whether the query succeeded.

brianhelba: I agree, when we consider folder move, we'd want to `SELECT FOR UPDATE` the ancestor rows within an explicit transaction which we open for the scope of the increment method. Until then, is there a need for a larger-scope transaction to tie these two queries (select the ancestor and update the sizes) together?

zachmullen: I think it's fine for now.

zachmullen (approved):

PR #60, opened 2020-10-26
Disallow changing folder's parent via REST API

zachmullen:
PTAL. Prior to this change, it was possible to alter a folder's parent reference via a REST update.

### DIFF
--- a/dkc/core/rest/folder.py
+++ b/dkc/core/rest/folder.py
@@ -16,6 +16,12 @@ class FolderSerializer(serializers.ModelSerializer):
         fields = ['id', 'name', 'description', 'parent', 'created', 'modified', 'size']
 
 
+class FolderUpdateSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Folder
+        fields = ['id', 'name', 'description']
+
+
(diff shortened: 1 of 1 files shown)


dkc/core/rest/folder.py
```diff
@@ -27,11 +33,15 @@ class FolderViewSet(ModelViewSet):
         AllowAny,
         # IsAuthenticatedOrReadOnly,
     ]
-    serializer_class = FolderSerializer
 
     filter_backends = [ActionSpecificFilterBackend]
     filterset_class = FoldersFilterSet
 
+    def get_serializer_class(self):
+        if self.request.method in ['PUT', 'PATCH']:
```
brianhelba, line 41:
```suggestion
        if self.action in ['update', 'partial_update']:
```

This is a more direct coupling to how DRF actually dispatches to the action methods which are the handlers.

brianhelba (changes requested):

brianhelba (approved):

zachmullen:
Not sure why CI is listed as "queued"... merging anyway since tests passed.

PR #61, opened 2020-10-26
Enforce max folder depth as a database constraint

brianhelba:
This pushes max folder depth enforcement to two extremes:
* In the model validation layer, where it's enforced via `full_clean` (or Form validations)
* In the database, where it's enforced as a fallback, which should typically never happen

This also starts the expectation that at Folder instantiation, the choice of `parent` will be decided (either `None` or specified). Changing a Folder's `parent` after instantiation confers a responsibility to update all associated denormalized fields (e.g. `depth`).


brianhelba:
I realized refactoring this was necessary as part of creating a Folder factory. It should be simpler to review as a separate change. I also think it's an improvement in its own right, as we now have fewer things happening via signals (which are hard to maintain) and more database constraints.

dkc/core/models/folder.py
```diff
@@ -8,23 +8,21 @@
 from django.dispatch import receiver
 from django_extensions.db.models import TimeStampedModel
 
-from dkc.core.exceptions import MaxFolderDepthExceeded
 from dkc.core.models.metadata import UserMetadataField
 
+MAX_DEPTH = 30
```
brianhelba, line 13: I had to make this module-scope. If it was an class attribute of `Folder`, it couldn't be referenced from the scope of `Folder.Meta`.

zachmullen (approved):

PR #64, opened 2020-10-29
Require creator field on File and Folder

brianhelba:


dkc/core/tests/conftest.py
```diff
@@ -6,10 +6,17 @@
 
 
 @pytest.fixture
-def api_client():
+def api_client() -> APIClient:
     return APIClient()
 
 
+@pytest.fixture
+def authenticated_api_client(user) -> APIClient:
```
zachmullen, line 14: I think this will have limited usefulness, since in most test cases we'd want a specific user that is linked via permissions to other documents, rather than an arbitrary user. But it's at least useful in some cases.

dkc/core/tests/test_file.py
```diff
@@ -20,7 +20,9 @@ def test_file_rest_retrieve(api_client, file):
 
 @pytest.mark.django_db
 def test_file_sibling_names_unique(file, file_factory):
-    sibling = file_factory.build(folder=file.folder, name=file.name, creator=None)
+    sibling = file_factory.build(folder=file.folder, name=file.name)
+    # Make sure foreign key references exist in database first
+    sibling.creator.save()
```
zachmullen, line 25: Can we just add the `user` fixture as a dependency and say `creator=user`? Seems a bit more elegant

brianhelba: I had it that way before, but it was conceptually helpful to distinguish factory parameters where are conceptually essential to the test case, from foreign key references which must be saved for purely technical reasons.

I liked this better, but I don't feel strongly.

dkc/core/tests/conftest.py
brianhelba: Until we have permissions, I think it will be useful. It may also be helpful for any testing of `create` endpoints, where the user doesn't have existing permissions on anything. I'd be fine with removing it later if it becomes obsolete.

brianhelba: I've removed this for now.

dkc/core/rest/folder.py
```diff
@@ -135,18 +135,19 @@ def get_serializer_context(self):
     @transaction.atomic
     def perform_create(self, serializer: serializers.ModelSerializer):
         parent: Folder = serializer.validated_data.get('parent')
```
brianhelba, line 137: The use of `FullCleanModelSerializer` is causing this to fail.

Read-only fields or non-serialized fields, like `creator`, are now required to be present on the input in order for `serializer.validated_data` to succeed.

Traditionally, `ModelViewSet.perform_create` (which runs after the serializer validates) is the right place to tack on additional mandatory, implicitly determined, data ([the DRF docs encourage this too](https://www.django-rest-framework.org/tutorial/6-viewsets-and-routers/#refactoring-to-use-viewsets)).

---

There is a solution: use `creator = ForeignKey(editable=False, ...)`. This causes `Model.full_clean()` to skip validation of that field (more precisely, [when the field validates itself, it is a no-op](https://github.com/django/django/blob/6822aa5c6c3fbec7c5393a05e990865ba59fe167/django/db/models/fields/__init__.py#L637)); this is the only reason that the `tree` field is able to work.

However, I don't think we want the `creator` to be non-editable (i.e. Django admins should be allowed to change this field, even if end users can't). As we develop other definitely-non-editable fields which require implicit creation, this problem will persist.

zachmullen (approved):

PR #66, opened 2020-10-29
Add object permissions based on django-guardian

jbeezley:
So far this only contains model layer logic on folders (files will be similar).  I will work on the rest layer next to give a better sense of how it is used.


dkc/core/models/tree.py
```diff
@@ -15,3 +19,23 @@ def root_folder(self) -> Folder:
...
+
+    def has_permission(self, user: User, permission: Permission) -> bool:
+        for perm in permission.derived_permissions:
+            if user.has_perm(perm, self):
+                return True
+        return False
+
+    class Meta:
+        permissions = (
+            (Permission.read.value.split('.')[1], 'Read access to a tree'),
```
zachmullen, line 38: Can the names not have `.` in them?

dkc/core/models/tree.py
```diff
@@ -15,3 +19,23 @@ def root_folder(self) -> Folder:
...
+        cls, user: User, permission: Permission, queryset: models.QuerySet['Tree']
+    ) -> models.QuerySet['Tree']:
+        perms = permission.derived_permissions
+        return get_objects_for_user(user, perms, klass=queryset, any_perm=True)
+
+    def has_permission(self, user: User, permission: Permission) -> bool:
+        for perm in permission.derived_permissions:
+            if user.has_perm(perm, self):
+                return True
+        return False
```
zachmullen, line 34: I think this is more readable / expressive as

```python
return any(user.has_perm(perm, self) for perm in permission.derived_permissions)
```

Also I find the name "derived_permissions" confusing here, as reading it out of context makes it seem backwards to what you expect. It's only the derived permissions if you're talking about *required* permission, not the permission *assigned* to a user. I'm struggling to come up with a great name, but I think capturing that it only works in that direction would be prudent.

dkc/core/tests/test_permissions.py
```diff
@@ -0,0 +1,71 @@
...
+@pytest.mark.django_db
+def test_tree_permissions_filter(user, no_access_tree, readable_tree, writeable_tree, admin_tree):
+    trees = Tree.filter_by_permission(user, Permission.read, Tree.objects).all()
+    assert {tree.id for tree in trees} == {admin_tree.id, writeable_tree.id, readable_tree.id}
+
+    trees = Tree.filter_by_permission(user, Permission.write, Tree.objects).all()
+    assert {tree.id for tree in trees} == {admin_tree.id, writeable_tree.id}
+
+    trees = Tree.filter_by_permission(user, Permission.admin, Tree.objects).all()
+    assert {tree.id for tree in trees} == {admin_tree.id}
```
zachmullen, line 48: This test seems like a good case for `@pytest.mark.parametrize`, no?

jbeezley:
I'll address  minor issues later, I'm pushing this as a proof of concept for the moment.  I also have to go through and fix the folder rest tests.

One thing to note about the behavior as implemented is that you get a 404 on requests for objects that you don't have read access to, but a 403 when trying to write to an object that you only have read access.  I could see an argument for wanting a 403 in both cases.

Finally, we need to decide on a method for providing `public` access either by a column on the tree table or by a default group.  I'm leaning toward the group method because I'm not sure how to add the condition inside the `has_perm` query.

zachmullen:
> One thing to note about the behavior as implemented is that you get a 404 on requests for objects that you don't have read access to, but a 403 when trying to write to an object that you only have read access. I could see an argument for wanting a 403 in both cases.

In my experience this is actually the most common behavior, and I'd say the most desirable for security. A 404 doesn't leak information about even the existence of a resource, other than perhaps via side channels.

From a behavioral perspective, it can be really nice for clients if, when read access is denied to an anonymous user, a 401 is returned. However, that subverts the "don't-leak-information-about-resource-existence" principle.

zachmullen:
> Finally, we need to decide on a method for providing public access either by a column on the tree table or by a default group. I'm leaning toward the group method because I'm not sure how to add the condition inside the has_perm query.

A group permission check seems like it requires a JOIN of at least 3 tables. It may be worth considering an optimized path like a `public` column on the Tree, which could obviate that database query entirely. I guess the decision comes down to how difficult it is to achieve that column check in our own code.

jbeezley:
> A group permission check seems like it requires a JOIN of at least 3 tables. It may be worth considering an optimized path like a public column on the Tree, which could obviate that database query entirely.

I realized I misspoke.  We can handle `has_perm` in python.  The issue is with integrating into `get_objects_for_user`.  It returns a [filtered queryset](https://github.com/django-guardian/django-guardian/blob/devel/guardian/shortcuts.py#L648), but we would need to find a way to add a special `OR` condition inside the filter.

jbeezley:
I'm to the point now where I need to design the rest interface for permissions.  I have two different options in mind:

1. Simple, something similar to:
```
POST /folders/:id/permissions permission=read user_id=1
DELETE /folders/:id/permissions permission=read user_id=1

POST /folders/:id/permissions permission=write group_id=1
DELETE /folders/:id/permissions permission=read group_id=1

GET /folders/:id/permissions
=>
{
  read: {users: [1, 4], groups: [1]},
  write: {users: [2], groups: []},
  admin: {users: [3], groups: []}
}
```

2. Elaborate--Full CRUD subroutes on `/folders/:id/permissions/users` and `/folders/:id/permissions/groups` using drf-nested-routers.

The other complication is what to do about permissions on users and groups themselves.  If we want user information to be private, then it begs the question of what to do about user permissions on a tree that the admin of the tree doesn't have access to.  Option 1 bypasses this for the most part by only returning the id of the users.  We would need to add `/users` and `/groups` routes to look up the actual user/group information in the UI, which could be completed in a follow-up PR.

jbeezley:
I added an MVP for the permissions api.  Individual permissions are serialized like `{'id': 1, 'model': 'user', 'permission': 'read'}`.  It's a little more verbose than what I described before, but it allows the permissions to be returned as a flat list.  I'm deferring an api for user group information for a future PR.

There are all sorts of edge cases we could add handling for e.g. preventing explicit permissions for admin users, preventing duplicate permissions for a single user, adding special semantics for public permissions, etc.

jbeezley:
This is almost ready for a more detailed review.  The only major thing I might want to change is swap out id's for group/user names in the permissions API.

dkc/core/rest/folder.py
```diff
@@ -39,22 +65,70 @@ class FoldersFilterSet(filters.FilterSet):
     parent = IntegerOrNullFilter(required=True)
 
 
+class FolderPermissionGrantSerializer(serializers.Serializer):
+    name = serializers.CharField(required=True)
```
zachmullen, line 69: Non-actionable comment for posterity: if we also wanted to support an ID-based API, this is forward compatible. We would just make `name` not required, and add an `id` field, and one of them would have to be set.

dkc/core/models/folder.py
```diff
@@ -116,6 +122,16 @@ def clean(self) -> None:
             )
         super().clean()
 
+    @classmethod
+    def filter_by_permission(
+        cls, user: User, permission: Permission, queryset: models.QuerySet['Folder']
+    ) -> models.QuerySet['Folder']:
+        tree_query = Tree.filter_by_permission(user, permission, Tree.objects).values('pk')
+        return queryset.filter(tree__in=models.Subquery(tree_query))
```
zachmullen, line 130: Another non-actionable comment for information. This builds queries of the form:

```python
>>> q = Folder.filter_by_permission(user, Permission.read, Folder.objects)
>>> print(q.query)
SELECT "core_folder"."id", "core_folder"."created", "core_folder"."modified", "core_folder"."name", "core_folder"."depth", "core_folder"."size", "core_folder"."description", "core_folder"."user_metadata", "core_folder"."parent_id", "core_folder"."tree_id" FROM "core_folder" WHERE "core_folder"."tree_id" IN ((SELECT "core_tree"."id" FROM "core_tree") UNION (SELECT "core_tree"."id" FROM "core_tree" WHERE "core_tree"."public")) ORDER BY "core_folder"."name" ASC

>>> q = Folder.filter_by_permission(user, Permission.write, Folder.objects)
>>> print(q.query)
SELECT "core_folder"."id", "core_folder"."created", "core_folder"."modified", "core_folder"."name", "core_folder"."depth", "core_folder"."size", "core_folder"."description", "core_folder"."user_metadata", "core_folder"."parent_id", "core_folder"."tree_id" FROM "core_folder" WHERE "core_folder"."tree_id" IN (SELECT U0."id" FROM "core_tree" U0) ORDER BY "core_folder"."name" ASC
```
```

File: dkc/core/rest/folder.py [hunk 6]
```diff
@@ -147,10 +220,102 @@ def agree_terms(self, request, pk=None):
...
+        for grant in grants:
+            tree.remove_permission(grant)
+
+    @transaction.atomic
+    def set_permission_list(self, tree: Tree, grants: List[PermissionGrant]):
+        self.grant_permission_list(tree, grants)
+
+        grant_set = set(grants)
+        existing_grant_set = set(tree.list_granted_permissions())
+        self.remove_permission_list(tree, list(existing_grant_set - grant_set))
```
zachmullen, line 255: This explains the `frozen`-ness of `PermissionGrant`. I like the set-based approach, but does it make sense to move all this stuff (`set_permission_list`, `remove_permission_list`, `grant_permission_list`) down into the Tree model?

zachmullen (changes requested):

dkc/core/permissions.py
```diff
@@ -0,0 +1,71 @@
...
+    permission: Permission
+
+
+class IsReadOnlyEndpoint(BasePermission):
+    def has_object_permission(self, request: Request, view: View, obj) -> bool:
+        return request.method in SAFE_METHODS
+
+
+class IsReadable(BasePermission):
+    def has_object_permission(self, request: Request, view: View, obj: Folder) -> bool:
```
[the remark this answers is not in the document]
jbeezley: I intended to implement a duck typed API for use with this class.  The minimal set of methods required are `filter_by_permission` and `has_permission`.  See [`e46203a`](https://github.com/girder/dkc-next/pull/66/commits/e46203a06d559c2dc9660c076bebd9448cff7152) where I implement it on files.

In general, this is what I had in mind for adding ACL's to additional models, but I wanted to get some feedback on it before writing elaborate documentation.

dkc/core/permissions.py
```diff
@@ -0,0 +1,71 @@
...
+    from dkc.core.models import Folder
+
+
+class Permission(Enum):
+    read = 'read_tree'
+    write = 'write_tree'
+    admin = 'admin_tree'
+
+    @property
+    def implied_permissions(self) -> List[str]:
```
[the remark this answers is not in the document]
jbeezley: I understand what you mean.  I think this is addressed in [`56dddf3`](https://github.com/girder/dkc-next/pull/66/commits/56dddf338e4131311d10dd138d0dccbfcee66440).

dkc/core/permissions.py
```diff
@@ -23,7 +23,16 @@ class Permission(Enum):
     admin = 'admin_tree'
 
     @property
-    def implied_permissions(self) -> List[str]:
+    def associated_permissions(self) -> List[str]:
```
zachmullen, line 26: 👍

zachmullen (approved):
LGTM. I'd like @brianhelba to also take a look, at least at the test code.

brianhelba (commented):
Some comments so far, I'm still looking at all of it...

mgrauer:
@jbeezley Can you please update this docs as part of this PR, and add all DKCN specific assumptions to the doc that drove your design?

https://github.com/girder/dkc-next/blob/master/docs/roadmap-authorization.md

jbeezley:
I added some documentation to the model/permissions layer.  I think it addresses the motivations for the code at a high level.  The roadmap doc can go into more detail as to policy decisions.

mgrauer:
@jbeezley what review do you need of this (after resolving merge conflicts) to get this merged? Is it code or comments or both?

jbeezley:
I guess mostly the comments.

mgrauer:
@brianhelba @zachmullen could one of you review comments in this PR?

dkc/core/models/tree.py
```diff
@@ -1,17 +1,168 @@
...
+            return set(Permission)
+
+        user_perms = get_user_perms(user, self)
+        return {Permission(p) for p in user_perms if p in permission_strings}
+
+    @transaction.atomic
+    def grant_permission(self, grant: PermissionGrant) -> None:
+        """Activate a specific permission grant.
+
+        This method with apply a user or group permission to the current entity.  This
```
zachmullen, line 92: with -> will

dkc/core/permissions.py
```diff
@@ -0,0 +1,103 @@
...
+
+if TYPE_CHECKING:
+    from dkc.core.models import File, Folder
+
+    Model = Union[File, Folder]
+
+
+class Permission(Enum):
+    """
+    A list tree access permissions.
```
zachmullen, line 22: This sentence doesn't parse for me, should it be "a list of"?

zachmullen (changes requested):
comments reviewed

dkc/core/models/tree.py
```diff
@@ -1,17 +1,168 @@
...
+        self.grant_permission_list(grants)
+
+        grant_set = set(grants)
+        existing_grant_set = set(self.list_granted_permissions())
+        self.remove_permission_list(list(existing_grant_set - grant_set))
+
+    def get_access(self, user: Optional[User]) -> Dict[str, bool]:
+        """Return the permissions that the given user has on this entity."""
+        if user is None:
+            # This is a special case for the anonymous user.  Acess can only
```
mgrauer, line 143:
```suggestion
            # This is a special case for the anonymous user.  Access can only
```

mgrauer (changes requested):

zachmullen (approved):

PR #67, opened 2020-10-30
Add basic tests for folder REST update

### LINKED ISSUE #63: Implement remaining Folder REST tests
Still need `test_folder_rest_update` and `test_folder_rest_destroy`. Both might need to be split into multiple cases, if we want to test behaviors for root / child or other varying circumstances.

* https://github.com/girder/dkc-next/blob/master/dkc/core/tests/test_folder_rest.py#L52
* https://github.com/girder/dkc-next/blob/master/dkc/core/tests/test_folder_rest.py#L58

zachmullen:
Fixes #63

### DIFF
--- a/dkc/core/tests/test_folder_rest.py
+++ b/dkc/core/tests/test_folder_rest.py
@@ -1,6 +1,6 @@
 import pytest
 
-from dkc.core.models import Folder
+from dkc.core.models import File, Folder, Tree
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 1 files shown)


zachmullen:
@brianhelba PTAL, this should be good to merge

dkc/core/tests/test_folder_rest.py
```diff
@@ -47,11 +47,24 @@ def test_folder_rest_retrieve(api_client, folder):
...
+    folder.refresh_from_db()
+    assert folder.name == 'New name'
+    assert folder.description == 'New description'
+
+
+@pytest.mark.django_db
+def test_folder_rest_update_parent_disallowed(api_client, folder, folder_factory):
+    other_root = folder_factory()
+    child = folder_factory(parent=folder)
+    api_client.patch(f'/api/v2/folders/{child.id}', data={'parent': other_root.id})
```
jbeezley, line 65: I would assert the status code returned is correct.

jbeezley (changes requested):

dkc/core/tests/test_folder_rest.py
```diff
@@ -62,7 +62,8 @@ def test_folder_rest_update(api_client, folder):
 def test_folder_rest_update_parent_disallowed(api_client, folder, folder_factory):
     other_root = folder_factory()
     child = folder_factory(parent=folder)
-    api_client.patch(f'/api/v2/folders/{child.id}', data={'parent': other_root.id})
+    resp = api_client.patch(f'/api/v2/folders/{child.id}', data={'parent': other_root.id})
+    assert resp.status_code == 200
```
jbeezley, line 66: I would expect that to be an error.  Is 200 the behavior dictated by DRF?

zachmullen: It is indeed DRF's behavior when trying to patch a non-editable field. It ignores it and continues on. I also would have hoped for a 400 here.

jbeezley: Apparently this is the default behavior of DRF when given an unknown field.  I wonder if making parent a [read_only_field](https://www.django-rest-framework.org/api-guide/serializers/#specifying-read-only-fields) would raise an error?

jbeezley: I guess it doesn't... oh well, I guess this is good enough.

jbeezley (approved):

PR #70, opened 2020-11-13
Enable usernames that are distinct from emails

jbeezley:


zachmullen (approved):
🧙🧙 Magic! 🧙🧙

zachmullen:
Are usernames restricted in any way? Like does this ensure that they won't be email addresses?

jbeezley:
The default validator is the regex `^[\w.@+-]+\Z`, so emails are allowed.

PR #71, opened 2020-11-16
Enforce constraints on usernames

zachmullen:


dkc/core/forms/signup.py
```diff
@@ -0,0 +1,26 @@
+import re
+
+from allauth.account.forms import SignupForm as SignupFormBase
+from django.core.exceptions import ValidationError
+
+username_regex = re.compile(r'^[a-z\d](?:[a-z\d]|-(?=[a-z\d])){0,49}$', flags=re.IGNORECASE)
+
+
+class SignupForm(SignupFormBase):
+    def clean_username(self):
```
brianhelba, line 10: I think this is a more appropriate place to run this logic: https://github.com/pennersr/django-allauth/blob/701d8ff905535e4cf79b6769ce9f60c7bed978ba/allauth/account/adapter.py#L249

We should be able to just add a callback to `ACCOUNT_USERNAME_VALIDATORS` ( https://django-allauth.readthedocs.io/en/latest/configuration.html ).

zachmullen: Excellent, I don't think @jbeezley or I were aware of that. I will change it.

zachmullen: @brianhelba I've converted it to use the new approach, but I'm not sure what the best way to test it is. Any thoughts?

zachmullen: Actually it turns out I can use the same approach and just use the out-of-the-box SignupForm from allauth.

zachmullen:
@brianhelba this is now ready for another review pass.

dkc/settings.py
```diff
@@ -10,6 +11,22 @@
     ProductionBaseConfiguration,
     TestingBaseConfiguration,
 )
+from django.core.validators import MaxLengthValidator, RegexValidator
+
+
+class UsernameRegexValidator(RegexValidator):
+    regex = r'^[a-z\d](?:[a-z\d]|-(?=[a-z\d]))*$'
```
brianhelba, line 18: Why subclass instead of passing these as arguments to the constructor?

zachmullen: I just copied how [these](https://github.com/django/django/blob/master/django/contrib/auth/validators.py) were done. I have no preference, I'm happy to change it if you want.

brianhelba (approved):

PR #96, opened 2020-12-15
Add OAuth and upgrade to cookiecutter v0.12

brianhelba:


setup.py
```diff
@@ -46,6 +46,7 @@
         'django-extensions',
         'django-filter',
         'django-girder-utils',
+        'django-oauth-toolkit==1.3.2',
```
zachmullen, line 49: Why the pinning?

zachmullen (approved):

setup.py
brianhelba: This pins one release before the latest. The latest release is broken (for the mode we're using it, Authorization Code with PKCE). It's fixed upstream in `master` now, but there hasn't been a new release.

See:
* https://github.com/jazzband/django-oauth-toolkit/issues/900
* https://github.com/girder/cookiecutter-django-girder/pull/95

PR #97, opened 2020-12-16
Switch to S3 file field

zachmullen:


brianhelba (approved):

mgrauer:
Was it really this painless? That's great if so.

brianhelba:
> Was it really this painless? That's great if so.

@mgrauer There are 3 more lines of code you need for a full install of S3FF, in `setup.py`, `settings.py`, `urls.py`, but those are already set up in the cookiecutter for us.

zachmullen:
@mgrauer [this](https://github.com/girder/girder_web_components/commit/86c1cc5e1808bfe456c5dc7d6e8b33684d15ab98) was the front-end change required.

mgrauer:
Good to know, thanks to you both for the info.

PR #101, opened 2021-01-07
Use JSONObjectField from django-girder-utils

brianhelba:


brianhelba:
This requires a minor upstream bug fix before use.

zachmullen:
This needs conflicts resolved, LGTM after that.

brianhelba:
@zachmullen Done.

zachmullen (approved):

PR #103, opened 2021-01-07
Ensure that all transitive requirements are installed

brianhelba:
The old pip resolver does not properly fetch transitive requirements in certain cases. Until the new pip resolver is available more consistently, this works around the bug.


jbeezley (approved):
That should do the trick.

Issue #73, opened 2020-11-16
Consider returning 404 status when no agreement to terms exists


brianhelba:
From #47

brianhelba:
Likewise with `user/me`.

zachmullen:
I don't have strong feelings, but I still prefer 204 because I think of these requests as valid even if the answer is a negative.

PR #109, opened 2021-01-08
Add endpoint for getting logged-in user's info

zachmullen:


dkc/core/rest/user.py
```diff
@@ -9,3 +13,17 @@ class Meta:
...
+class UserViewSet(GenericViewSet):
+    queryset = User.objects.all()
+
+    permission_classes = [AllowAny]
+
+    @action(detail=False, pagination_class=None)
+    def me(self, request):
+        """Return the currently logged in user's information."""
+        if request.user.is_anonymous:
+            return Response(status=204)
```
brianhelba, line 27: IMO, `404` is more appropriate, but I think we should defer that bikeshedding to #73.

dkc/core/rest/user.py
```diff
@@ -9,3 +13,17 @@ class Meta:
             'id',
             'username',
         ]
+
+
+class UserViewSet(GenericViewSet):
+    queryset = User.objects.all()
+
+    permission_classes = [AllowAny]
```
brianhelba, line 21: In the future, we may want the user list to not be available to all, but this is fine for now.

brianhelba (approved):
LGTM.

dkc/core/rest/user.py
brianhelba: Nevermind, I didn't realize it was a `GenericViewSet`, and doesn't provide any default endpoints.

PR #110, opened 2021-01-12
Perform validation on files in REST layer

zachmullen:
@brianhelba PTAL, my goal here is to ensure that the REST layer gives 400 errors with reasonable messages when model create/update fails validation. I have no idea where is the best place to put this validation logic.

### DIFF
--- a/dkc/core/rest/utils.py
+++ b/dkc/core/rest/utils.py
@@ -1,0 +1,19 @@
+from django.core.exceptions import ValidationError as DjangoValidationError
+from rest_framework import serializers
+
+
+class FullCleanModelSerializer(serializers.ModelSerializer):
+    def validate(self, data):
+        if self.partial:  # partial update
+            for k, v in data.items():
+                setattr(self.instance, k, v)
+            try:
+                self.instance.full_clean()
+            except DjangoValidationError as exc:
+                raise serializers.ValidationError(serializers.as_serializer_error(exc))
+        else:  # create or full update
+            try:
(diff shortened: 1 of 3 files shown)


zachmullen:
Nobody seems to have a great answer for where to do validation... if you know of a better way, I'd love to hear it.

dkc/core/rest/utils.py
```diff
@@ -0,0 +1,19 @@
+from django.core.exceptions import ValidationError as DjangoValidationError
+from rest_framework import serializers
+
+
+class FullCleanModelSerializer(serializers.ModelSerializer):
+    def validate(self, data):
+        if self.partial:  # partial update
+            for k, v in data.items():
+                setattr(self.instance, k, v)
```
brianhelba, line 9: We're mutating the only copy of `self.instance` here, which means that validation (even if unsuccessful) has side effects on the existing model. If a request doesn't terminate immediately after this, we'll end up with an invalid model, which could even be a problem for debugging.

zachmullen: > If a request doesn't terminate immediately after this, we'll end up with an invalid model, which could even be a problem for debugging.

Do you mean to say that a model instance object may be shared across different requests...?

zachmullen: Ignore deleted comment, I misread your last sentence.

brianhelba (approved):
Seems reasonable for now. I hope that as our business logic becomes more complex, it will be more obvious where the most natural place for it is.

PR #119, opened 2021-01-18
Permission-related bugfixes

zachmullen:


dkc/core/rest/folder.py
```diff
@@ -134,7 +136,17 @@ def get_serializer_context(self):
     @transaction.atomic
     def perform_create(self, serializer: serializers.ModelSerializer):
         parent: Folder = serializer.validated_data.get('parent')
-        tree = parent.tree if parent else Tree.objects.create()
+        if parent:
+            tree = parent.tree
+            if not tree.has_permission(serializer.context['user'], permission=Permission.write):
+                raise PermissionDenied()
```
jbeezley, line 142: Not certain, but we may need something similar in the file creation endpoint.

jbeezley: NVM, I just saw it.

dkc/core/rest/file.py
```diff
@@ -31,12 +33,18 @@ class Meta:
 class FileViewSet(ModelViewSet):
     queryset = File.objects.all()
 
-    permission_classes = [HasAccess]
+    permission_classes = [HasAccess & IsAuthenticatedOrReadOnly]
```
jbeezley, line 36: I believe `IsAuthenticatedOrReadOnly` logic is redundant with the `perform_create` checks as it is not possible to give the "anonymous user" write access to anything.  I think if we wanted to make this explicit we should change `HasAccess` to
```python
HasAccess = (IsReadOnlyEndpoint & IsReadable) | (IsWriteable & IsAuthenticated)
```

dkc/core/rest/file.py
```diff
@@ -31,12 +33,18 @@ class Meta:
...
     serializer_class = FileSerializer
 
     filter_backends = [PermissionFilterBackend, ActionSpecificFilterBackend]
     filterset_fields = ['folder', 'sha512']
 
+    def perform_create(self, serializer: FileSerializer):
+        folder: Folder = serializer.validated_data.get('folder')
+        if not folder.has_permission(self.request.user, permission=Permission.write):
+            raise PermissionDenied()
+        serializer.save()
```
jbeezley, line 46: This contradicts what I said this morning, but I now realize this logic probably belongs in a permission class to be consistent with how the other permissions are enforced.  I'm thinking perhaps to add an explicit API for object creation on the model classes.

Something like this:
```python
class Cls:
	@classmethod
    def can_create(cls, user, obj) -> bool:
		# return true if the user has permission to create the object
```

Then, a generic permissions class that calls that method.  The thing I'm not sure about is how to create the object inside permission class.  I don't know if you can get the serializer from the request object or not...  Alternatively, we could pass `request.data` into that method.

If this gets too hairy, I think we can save it for later because it's an important bug fix.

jbeezley (changes requested):

dkc/core/rest/file.py
zachmullen: I'm happy to make that change. The reason I added this here is because it causes the file creation endpoint to return 401 for anonymous users, which is desired, whereas without it it returns 403 because it falls back to object-level permission checks.

dkc/core/rest/file.py
zachmullen: I'm all for doing this the right way, though I'd rather have something that works in place, and I don't think I'm personally familiar enough with the best practices here to make an attempt at your suggestion. Maybe we can merge this and open an issue to move it.

jbeezley: Works for me.  I can make an attempt at it.

dkc/core/rest/file.py
jbeezley: That makes sense.  I think I would prefer it on the default permissions class rather than added everywhere it's used.

zachmullen: I changed it to 
```python
HasAccess = (IsReadOnlyEndpoint & IsReadable) | (IsAuthenticated & IsWriteable)
```

But that gives 403s instead of 401s. I will try and figure out why that's happening.

zachmullen: The explanation is [here](https://www.django-rest-framework.org/api-guide/permissions/#how-permissions-are-determined)

zachmullen: Nevermind, that was a red herring. The real issue was that `IsReadOnlyEndpoint` was using `has_object_permission` but it really needs to use `has_permission`, which also gets called on create endpoints.

jbeezley: Ah.  That makes sense.

zachmullen:
Ready for final review.

dkc/core/tests/conftest.py
```diff
@@ -28,6 +29,21 @@ def public_folder(folder_factory):
...
+
+    Because S3FF currently doesn't check that the object was created, this just
+    provides a fake signed value rather than mocking the entire S3FF upload process.
+    """
+    return signing.dumps(
+        {
+            'object_key': 'key',
+            'file_size': 123,
+        }
+    )
```
zachmullen, line 44: FYI @brianhelba

jbeezley (approved):

PR #120, opened 2021-01-18
Set serializer class on UserViewSet

zachmullen:
This addresses an assertion error I noticed on Sentry:

```'UserViewSet' should either include a `serializer_class` attribute, or override the `get_serializer_class()` method.```

### DIFF
--- a/dkc/core/rest/user.py
+++ b/dkc/core/rest/user.py
@@ -19,6 +19,7 @@ class UserViewSet(GenericViewSet):
     queryset = User.objects.all()
 
     permission_classes = [AllowAny]
+    serializer_class = UserSerializer
 
     @action(detail=False, pagination_class=None)
     def me(self, request):

### CODE REVIEW

File: dkc/core/rest/user.py [hunk 1]
```diff
@@ -19,6 +19,7 @@ class UserViewSet(GenericViewSet):
     queryset = User.objects.all()
 
     permission_classes = [AllowAny]
+    serializer_class = UserSerializer
```
brianhelba, line 22: You could update line 29 too, to get the serializer from the surrounding ViewSet, but maybe you don't want to do it for actions.

brianhelba (approved):

dkc/core/rest/user.py
zachmullen: I prefer using the class explicitly.

PR #121, opened 2021-01-20
Allow explicit permissions to be added for superusers

jbeezley:
This fixes something that could be seen as a bug where permissions weren't actually being added for superusers.  There is no behavioral change here because superusers can always do anything anyway.

Alternatively, we could make it so superusers are always included in the list of permissions returned by `GET /folders/{id}/permissions`.  This has the problem that the implicit super user permissions couldn't be removed, and it would expose a global list of superusers to everyone.


zachmullen (approved):
Thanks, I like this better. I think it was confusing when you added a user to the permission list and saved it successfully, but then they didn't appear in the permission list after you refreshed.

PR #126, opened 2021-01-21
Use pytest-randomly to set a consistent random seed

jbeezley:
Should prevent the random testing errors that we get due to name constraints.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -50,6 +50,7 @@ deps =
     pytest-django
     pytest-factoryboy
     pytest-mock
+    pytest-randomly
 commands =
     pytest {posargs}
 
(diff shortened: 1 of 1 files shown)


zachmullen (approved):
Seems fine to me, unless someone finds value in using slightly different values for fields each time. It's not fuzzing per se, so I don't really think it matters.

jbeezley:
You can use `-p no:randomly` if you really want no seed.

brianhelba:
I think a lot of the benefit just comes from having values which we didn't choose, which prevents some amount of overfitting to test cases. Additionally, having a large number of different values is nice for the same reasons, and we do loose that aspect a bit here.

The alternative to this change is to update fields which are in danger of collision to use a value of a `Faker` + `Sequence`, which will prevent collisions but give us the full diversity of the random pool across other fields.

brianhelba:
I'd advocate my approach (`Faker` + `Sequence`), but I'm open to others' thoughts. I imagine that most fields can remain only with `Faker`, it's just the unique + large cardinality ones that'd need to be changed.

PR #130, opened 2021-01-30
Rework FolderSerializer to use native DRF validation

brianhelba:
`FullCleanModelSerializer` does not work properly with internally-set fields like `tree`, since the value is not available at the time of initial validation. Currently, it only works because `tree` has `editable=False`, so the Model-layer validator skips the field.

See [this comment](https://github.com/girder/dkc-next/pull/64#discussion_r567090441) for more information on the problem.

This refactor also has the result of effectively removing the default of `parent=None` when creating a folder. This is probably a net benefit, as it forces API clients to be explicit when creating a new root folder.

Also:
* Add tests for POSTing invalid Folder values
* Test folder updates with a PATCH request

### DIFF
--- a/dkc/core/rest/folder.py
+++ b/dkc/core/rest/folder.py
@@ -47,14 +47,14 @@ class FolderSerializer(serializers.ModelSerializer):
             serializers.UniqueTogetherValidator(
                 queryset=Folder.objects.all(),
                 fields=['parent', 'name'],
-                message='Folder names must be unique among siblings.',
+                message='A folder with that name already exists here.',
             ),
             # This could also be implemented as a UniqueValidator on 'name',
             # but its easier to not explicitly redefine the whole serializer field
             serializers.UniqueTogetherValidator(
                 queryset=Folder.objects.filter(parent=None),
                 fields=['name'],
-                message='Root folders must have a unique name.',
+                message='A root folder with that name already exists.',
             ),
             # folder_max_depth and unique_root_folder_per_tree are internal sanity constraints,
             # and do not need to be enforced as validators
--- a/dkc/core/rest/folder.py
+++ b/dkc/core/rest/folder.py
@@ -12,7 +12,7 @@ from rest_framework.exceptions import ValidationError
 from rest_framework.response import Response
 from rest_framework.viewsets import ModelViewSet
 
-from dkc.core.models import Folder, Terms, TermsAgreement, Tree
+from dkc.core.models import File, Folder, Terms, TermsAgreement, Tree
 from dkc.core.permissions import (
     HasAccess,
     IsAdmin,
@@ -62,6 +62,26 @@ class FolderSerializer(serializers.ModelSerializer):
 
     def get_access(self, folder: Folder) -> Dict[str, bool]:
         return folder.tree.get_access(self.context.get('user'))
+
+    def validate(self, attrs):
+        self._validate_unique_file_siblings(attrs)
+        return attrs
+
+    def _validate_unique_file_siblings(self, attrs):
+        if self.instance is None:
+            # Create
+            # By this point, other validators will have run, ensuring that 'name' and 'parent' exist
+            name = attrs['name']
+            parent_id = attrs['parent']
+        else:
+            # Update
+            # On a partial update, 'name' and 'parent' might be absent, so use the existing instance
+            name = attrs['name'] if 'name' in attrs else self.instance.name
+            parent_id = attrs['parent_id'] if 'parent_id' in attrs else self.instance.parent_id
+        if parent_id is not None and File.objects.filter(name=name, folder_id=parent_id).exists():
+            raise serializers.ValidationError(
+                'A file with that name already exists here.', code='unique'
+            )
 
 
 class FolderUpdateSerializer(FolderSerializer):
--- a/dkc/core/rest/folder.py
+++ b/dkc/core/rest/folder.py
@@ -23,10 +23,9 @@ from dkc.core.permissions import (
 )
 
 from .filtering import ActionSpecificFilterBackend, IntegerOrNullFilter
-from .utils import FullCleanModelSerializer
 
 
-class FolderSerializer(FullCleanModelSerializer):
+class FolderSerializer(serializers.ModelSerializer):
     public: bool = serializers.BooleanField(read_only=True)
     access: Dict[str, bool] = serializers.SerializerMethodField()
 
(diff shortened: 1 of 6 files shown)


dkc/core/rest/folder.py
```diff
@@ -43,14 +42,31 @@ class Meta:
...
+                message='Folder names must be unique among siblings.',
+            ),
+            # This could also be implemented as a UniqueValidator on 'name',
+            # but its easier to not explicitly redefine the whole serializer field
+            serializers.UniqueTogetherValidator(
+                queryset=Folder.objects.filter(parent=None),
+                fields=['name'],
+                message='Root folders must have a unique name.',
+            ),
+            # folder_max_depth and unique_root_folder_per_tree are internal sanity constraints,
```
brianhelba, line 59: Actually, `folder_max_depth` is something we might want to check at the REST layer (rather than allowing the database to do it). I'm not sure how to do this, so I'll probably file a bug for a future fix if nobody else has suggestions.

dkc/core/rest/folder.py
```diff
@@ -43,14 +42,31 @@ class Meta:
...
+        ]
 
     def get_access(self, folder: Folder) -> Dict[str, bool]:
         return folder.tree.get_access(self.context.get('user'))
 
 
 class FolderUpdateSerializer(FolderSerializer):
     class Meta(FolderSerializer.Meta):
-        fields = ['id', 'name', 'description']
+        read_only_fields = ['parent']
```
brianhelba, line 69: Specifying `fields` had the result of also limiting the *output* of what this serializer produced in the response to a `PUT`/`PATCH` request. What we really want to do is ensure that 'parent' cannot be changed in an update request, which this now does.

dkc/core/rest/folder.py
```diff
@@ -134,16 +150,17 @@ def get_serializer_context(self):
     # Atomically roll back the tree creation if folder creation fails
     @transaction.atomic
     def perform_create(self, serializer: serializers.ModelSerializer):
-        parent: Folder = serializer.validated_data.get('parent')
+        parent: Folder = serializer.validated_data['parent']
```
brianhelba, line 153: `parent` is not optional, so we should rely on it being there.

dkc/core/rest/folder.py
```diff
@@ -134,16 +150,17 @@ def get_serializer_context(self):
...
+        user: User = self.request.user
         if parent:
             tree = parent.tree
-            if not tree.has_permission(serializer.context['user'], permission=Permission.write):
```
brianhelba, line 140: Serializer context seems to be intended to provide additional data *into* the serializer, and not for pulling other data *out* of it (that's the role of `HiddenField`).

In this method, we have more direct access to the authoritative `self.request.user` anyway.

dkc/core/rest/folder.py
```diff
@@ -43,14 +42,31 @@ class Meta:
             'public',
             'access',
         ]
+        # ModelSerializer cannot auto-generate validators for model-level constraints
+        validators = [
+            serializers.UniqueTogetherValidator(
+                queryset=Folder.objects.all(),
+                fields=['parent', 'name'],
+                message='Folder names must be unique among siblings.',
```
zachmullen, line 50: This error message is a little technical -- keep in mind, in the SPA I just display these messages as-is to end users. How about this instead?

"A folder with that name already exists here."

dkc/core/rest/folder.py
```diff
@@ -43,14 +42,31 @@ class Meta:
...
+            ),
+            # This could also be implemented as a UniqueValidator on 'name',
+            # but its easier to not explicitly redefine the whole serializer field
+            serializers.UniqueTogetherValidator(
+                queryset=Folder.objects.filter(parent=None),
+                fields=['name'],
+                message='Root folders must have a unique name.',
```
zachmullen, line 57: This could be the same message I suggested above if that's fine with you.

dkc/core/tests/test_folder_rest.py
```diff
@@ -51,7 +95,7 @@ def test_folder_rest_retrieve(admin_api_client, folder):
 
 @pytest.mark.django_db
 def test_folder_rest_update(admin_api_client, folder):
-    resp = admin_api_client.put(
+    resp = admin_api_client.patch(
```
brianhelba, line 98: We might want to outright disallow `PUT` requests at some point, but since I think we typically expect clients to use `PATCH` requests, it's better for this test to use that.

zachmullen (changes requested):
Could we expand the scope of this PR to deleting `FullCleanModelSerializer` altogether? Based on this PR and our discussion, it seems like we shouldn't be using it anywhere. I think the only other changes needed would be in rest/file.py.

brianhelba:
@zachmullen I intend to submit a second PR to remove `FullCleanModelSerializer` entirely, but it's going to be more manageable to refactor `FileSerializer` separately.

dkc/core/rest/folder.py
```diff
@@ -43,14 +42,31 @@ class Meta:
...
+            ),
+            # folder_max_depth and unique_root_folder_per_tree are internal sanity constraints,
+            # and do not need to be enforced as validators
+        ]
```
zachmullen, line 61: It looks like this doesn't yet capture the logic that's in `Folder.clean`: checking that there's also no `File` with the same name.

brianhelba:
Tests are failing due to https://github.com/girder/django-composed-configuration/pull/114 . They'll pass as soon as that's released.

dkc/core/rest/folder.py
brianhelba: You're right, I missed that. I'll try to fix it and add a test.

zachmullen (changes requested):
After this change, the error response has the form:

```json
{
  "non_field_errors": [
    "..."
  ]
}
```

I think we should associate these UniqueTogetherValidators with the `name` field, as that will be more consumable downstream. (Is that difficult to do?)

brianhelba:
> After this change, the error response has the form:
> 
> ```json
> {
>   "non_field_errors": [
>     "..."
>   ]
> }
> ```
> 
> I think we should associate these UniqueTogetherValidators with the `name` field, as that will be more consumable downstream. (Is that difficult to do?)

For the case of sibling files, this is easy, as we just need to explicitly `raise ValidationError({'name': ...})`, and DRF will skip nesting it under [`NON_FIELD_ERRORS_KEY`](https://www.django-rest-framework.org/api-guide/settings/#non_field_errors_key).

In the cases where `UniqueTogetherValidator` is used, setting `message` to be a dict instead of a string causes an error, as [DRF calls `.format` on the `message` value](https://github.com/encode/django-rest-framework/blob/19655edbf782aa1fbdd7f8cd56ff9e0b7786ad3c/rest_framework/validators.py#L159). I've created a `FormattableDict` to wrap the values as a workaround.

zachmullen (approved):

zachmullen:
Let's open that ticket for validating max depth.

brianhelba:
Pushed 2 more minor commits.

brianhelba:
As a note, following the advice to `print(repr(FolderSerializer()))` from https://www.django-rest-framework.org/api-guide/validators/#validation-in-rest-framework was quite helpful. Given how much of `ModelSerializer` subclasses are auto-generated, future development should always include this as an essential sanity check.

PR #132, opened 2021-01-31
Rework FileSerializer to use native DRF validation

brianhelba:


zachmullen (approved):
Thanks for leading this charge, LGTM

PR #133, opened 2021-02-01
Add user-based quota enforcement

### LINKED ISSUE #76: Add quotas to folder roots

brianhelba:
This does not yet support "project-based" quotas. That will be added in a future PR, as it requires a UI to manage it.

This supersedes PRs #41, #49, #50.

Fixes #76.


dkc/core/models/quota.py
```diff
@@ -1,26 +1,62 @@
...
+                check=models.Q(used__lte=models.F('allowed')),
+                name='used_lte_allowed',
+            )
+        ]
+
+    used = models.PositiveBigIntegerField(default=0)
+    allowed = models.PositiveBigIntegerField(default=_default_user_quota)
+
+    # OneToOneField ensures that only one Quota per User exists
+    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name='quota')
```
brianhelba, line 28: Eventually, with "project-based" quotas, this can be nullable.

Ideally, the relationship would point the other way, but since `User` is practically not directly extensible, this is a reasonable substitute. Given the `_create_user_quota` signal handler, we can rely on `user.quota` always existing for now.

dkc/core/tests/test_folder.py
```diff
@@ -87,11 +85,9 @@ def test_folder_sibling_names_unique_files(file, folder_factory):
 
 @pytest.mark.django_db
 def test_root_folder_names_unique(folder, folder_factory):
-    other_root = folder_factory.build(name=folder.name)
```
brianhelba, line 90: As the models' foreign key relationships have gotten more complex, the factory "build" strategy is becoming more and more unwieldy. I plan to introduce a follow-up PR to remove nearly all usage of the "build" strategy.

dkc/settings.py
```diff
@@ -40,6 +40,8 @@ class DkcMixin(ConfigMixin):
     ACCOUNT_USER_MODEL_USERNAME_FIELD = 'username'
     ACCOUNT_AUTHENTICATION_METHOD = 'username_email'
 
+    DKC_DEFAULT_QUOTA = 3 << 30  # 3 GB
```
brianhelba, line 43: Is this a reasonable default?

dkc/core/rest/folder.py
```diff
@@ -182,7 +182,8 @@ def perform_create(self, serializer: FolderSerializer):
             if not tree.has_permission(user, permission=Permission.write):
                 raise PermissionDenied()
         else:
-            tree = Tree.objects.create()
+            # Use the user's quota for the new tree
+            tree = Tree.objects.create(quota=user.quota)
```
brianhelba, line 186: Just a note at how powerful this particular line is. I'm very pleased at how easy this was.

dkc/core/admin/quota.py
```diff
@@ -0,0 +1,42 @@
+from django.contrib import admin
+from django.db import models
+from django_admin_display import admin_display
+import humanize
+
+from dkc.core.models import Quota
+
+
+@admin.register(Quota)
+class QuotaAdmin(admin.ModelAdmin):
```
brianhelba, line 10: Note to QA, please take a look at the running admin page.

dkc/core/models/quota.py
```diff
@@ -1,26 +1,62 @@
...
+            return
 
-    # The user is needed to stringify, so join it when directly querying for a quota
-    objects = SelectRelatedManager('user')
+        try:
+            # Use an .update query instead of a .save, to avoid assigning an F-expression on the
+            # local instance, which might need to be rolled back on a failure
+            Quota.objects.filter(pk=self.pk).update(used=(models.F('used') + amount))
+        except IntegrityError as e:
+            if '"used_lte_allowed"' in str(e):
```
zachmullen, line 48: I wish we didn't have to do this -- do you think we'd have any luck upstreaming a feature request to psycopg2 to provide more structured info on their IntegrityErrors?

dkc/core/models/quota.py
zachmullen: Could we pull this out of here and have a separate associative table to map users to quotas? Then, every user could have an entry in that table, but not every quota would have to have an entry there (i.e. for any non-user associated quotas).

dkc/core/admin/quota.py
zachmullen: I took a look at the admin page, but it is broken for me on loading the quotas page:

```
Environment:


Request Method: GET
Request URL: http://localhost:8000/admin/core/quota/

Django Version: 3.1.6
Python Version: 3.8.7
Installed Applications:
['django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.sites',
 'girder_style',
 'allauth',
 'allauth.account',
 'allauth.socialaccount',
 'django.contrib.postgres',
 'whitenoise.runserver_nostatic',
 'django.contrib.staticfiles',
 'corsheaders',
 'django_filters',
 'rest_framework',
 'rest_framework.authtoken',
 'oauth2_provider',
 'drf_yasg',
 'django_extensions',
 'debug_toolbar',
 'dkc.core.apps.CoreConfig',
 's3_file_field',
 'guardian']
Installed Middleware:
['debug_toolbar.middleware.DebugToolbarMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware']



Traceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/contrib/admin/options.py", line 614, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/contrib/admin/sites.py", line 233, in inner
    return view(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/utils/decorators.py", line 130, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/contrib/admin/options.py", line 1693, in changelist_view
    cl = self.get_changelist_instance(request)
  File "/usr/local/lib/python3.8/site-packages/django/contrib/admin/options.py", line 735, in get_changelist_instance
    return ChangeList(
  File "/usr/local/lib/python3.8/site-packages/django/contrib/admin/views/main.py", line 99, in __init__
    self.queryset = self.get_queryset(request)
  File "/usr/local/lib/python3.8/site-packages/django/contrib/admin/views/main.py", line 450, in get_queryset
    ) = self.get_filters(request)
  File "/usr/local/lib/python3.8/site-packages/django/contrib/admin/views/main.py", line 152, in get_filters
    spec = field_list_filter_class(
  File "/usr/local/lib/python3.8/site-packages/django/contrib/admin/filters.py", line 435, in __init__
    raise ImproperlyConfigured(

Exception Type: ImproperlyConfigured at /admin/core/quota/
Exception Value: The list filter 'EmptyFieldListFilter' cannot be used with field 'user' which doesn't allow empty strings and nulls.
```
Request Method: | GET
-- | --
http://localhost:8000/admin/core/quota/
3.1.6
ImproperlyConfigured
The list filter 'EmptyFieldListFilter' cannot be used with field 'user' which doesn't allow empty strings and nulls.
/usr/local/lib/python3.8/site-packages/django/contrib/admin/filters.py, line 435, in __init__
/usr/local/bin/python
3.8.7
['/opt/django-project',  '/opt/django-project',  '/usr/local/lib/python38.zip',  '/usr/local/lib/python3.8',  '/usr/local/lib/python3.8/lib-dynload',  '/usr/local/lib/python3.8/site-packages',  '/opt/django-project']


Request Method: | GET
-- | --
http://localhost:8000/admin/core/quota/
3.1.6
ImproperlyConfigured
The list filter 'EmptyFieldListFilter' cannot be used with field 'user' which doesn't allow empty strings and nulls.
/usr/local/lib/python3.8/site-packages/django/contrib/admin/filters.py, line 435, in __init__
/usr/local/bin/python
3.8.7
['/opt/django-project',  '/opt/django-project',  '/usr/local/lib/python38.zip',  '/usr/local/lib/python3.8',  '/usr/local/lib/python3.8/lib-dynload',  '/usr/local/lib/python3.8/site-packages',  '/opt/django-project']

dkc/core/models/quota.py
zachmullen: (Not to hold up this PR or anything, just an idle thought.)

dkc/core/models/quota.py
brianhelba: I think that would be a `Profile` model ([Option 2 here](https://simpleisbetterthancomplex.com/tutorial/2016/07/22/how-to-extend-django-user-model.html)), which would also store any other user-related info that we wanted to extend.

Do you foresee any upcoming features where we want to store additional information (beyond just foreign key relationships) with the user? If so, I think it's worth a detour to add the `Profile` now.

If not, I think it's worth merging this approach as-is, to start giving us working quota support. As we extend to non-user-associated ("project") quotas, we can rework the implementation internally.

zachmullen: Let's add the profile model, it's not much code and should be straightforward. Just make it a commit on this same topic.

dkc/core/models/quota.py
brianhelba: [Here's where the `IntegrityError` is generated.](https://github.com/psycopg/psycopg2/blob/f469331af5f7b236f8037b302b69c2bf0df2a096/psycopg/error_type.c#L81-L82)

[The PostgreSQL docs state](https://www.postgresql.org/docs/current/errcodes-appendix.html):
> For some types of errors, the server reports the name of a database object (a table, table column, data type, or constraint) associated with the error; for example, the name of the unique constraint that caused a unique_violation error. Such names are supplied in separate fields of the error report message so that applications need not try to extract them from the possibly-localized human-readable text of the message. As of PostgreSQL 9.3, complete coverage for this feature exists only for errors in SQLSTATE class 23 (integrity constraint violation), but this is likely to be expanded in future.

So, it looks like it's possible in principle for only this particular case. [Psycopg 2 supports >= PostgreSQL 7.4](https://www.psycopg.org/docs/install.html#prerequisites), so support would need to be conditional.

@zachmullen I would start by filing an issue [in Psycopg 2](https://github.com/psycopg/psycopg2/issues), but if they decline it there, perhaps they'd consider it in (the apparently actively-developed) [Psycopg 3](https://github.com/psycopg/psycopg3/issues). Let me know if you do so, so I can subscribe to it.

dkc/core/admin/quota.py
brianhelba: Fixed.

dkc/core/admin/quota.py
```diff
@@ -0,0 +1,39 @@
...
+
+    readonly_fields = ['used', 'usage_percent']
+
+    @admin_display(short_description='Used', admin_order_field='used')
+    def human_used(self, obj: Quota) -> str:
+        return humanize.naturalsize(obj.used)
+
+    @admin_display(short_description='Allowed', admin_order_field='used')
+    def human_allowed(self, obj: Quota) -> str:
+        return humanize.naturalsize(obj.allowed)
```
zachmullen, line 24:
```suggestion
        return humanize.naturalsize(obj.allowed, binary=True)
```

dkc/core/admin/quota.py
```diff
@@ -0,0 +1,39 @@
...
+    list_display = ['id', 'user', 'human_used', 'human_allowed', 'usage_percent']
+    list_select_related = ['user']
+
+    search_fields = ['user__username']
+
+    readonly_fields = ['used', 'usage_percent']
+
+    @admin_display(short_description='Used', admin_order_field='used')
+    def human_used(self, obj: Quota) -> str:
+        return humanize.naturalsize(obj.used)
```
zachmullen, line 20:
```suggestion
        return humanize.naturalsize(obj.used, binary=True)
```

zachmullen (changes requested):

dkc/core/models/quota.py
brianhelba: I started to create a `Profile` model. This is something that's supported by Factory Boy: https://factoryboy.readthedocs.io/en/latest/recipes.html#example-django-s-profile , but a significant complication is the fact that we'll have to annotate all of our factories (not just `UserFactory` and `ProfileFactory`) with `@factory.django.mute_signals(post_save)`, which will disable any `post_save` handlers during essentially all tests. Eventually, we might need this, but it will introduce some drag.

IMO, the current implementation is adequate. At runtime, the reverse relation `user.quota` is adequate to support this feature and any lookup failures will immediately log 500 errors. If we can't depend on signals to support some behavior (instead requiring DB-level guarantees), I think we'd be giving up an important measure of power that Django provides.

In some ways, attaching the `ForeignKey` to `Quota` is superior, particularly once that field is nullable (to support project quotas) in the future, since we'll be able to easily filter user quotas from project quotas (I'm not seeing how to do this if the `ForeignKey` reference points *to* `Quota`).

Given the downside of suppressing all `post_save` signals, I think we should only proceed with `Profile` if we are going to need it to store additional information on each user. Otherwise, I'd advocate to merge this as-is, and proceed with developing more advanced features on top.

zachmullen: I just want to be clear that putting the user field here is a relational hack. We're misusing nullability to model what should be a relation. We can merge this as is, but I'd like to create an issue to address this. Insofar as this repository is meant to model best practices, one of the things I really want us to exemplify is proper use of relational databases with Django, which means we'll need to work through the warts. As nice as FactoryBoy is, we can't let it be the library driving our design choices.

Let's make the ticket, I'll take a crack at this myself later on.

zachmullen (approved):

PR #137, opened 2021-02-03
Allow folder creation through django admin

zachmullen:
FYI @danlamanna

### DIFF
--- a/dkc/core/admin/folder.py
+++ b/dkc/core/admin/folder.py
@@ -1,6 +1,7 @@
 from django.contrib import admin
+from django.db import transaction
 
-from dkc.core.models import Folder
+from dkc.core.models import Folder, Tree
 
 
 @admin.register(Folder)
@@ -9,3 +10,12 @@ class FolderAdmin(admin.ModelAdmin):
     list_display_links = ['id', 'name']
 
     search_fields = ['name', 'id']
+
+    @transaction.atomic
+    def save_model(self, request, obj: Folder, form, change: bool):
+        if not change:
+            if obj.parent:
+                obj.tree = obj.parent.tree
+            else:
+                obj.tree = Tree.objects.create(quota=obj.creator.quota)
+        super().save_model(request, obj, form, change)


dkc/core/admin/folder.py
```diff
@@ -9,3 +11,15 @@ class FolderAdmin(admin.ModelAdmin):
...
 
     search_fields = ['name', 'id']
+
+    @transaction.atomic
+    def save_model(self, request, obj: Folder, form, change: bool):
+        if not change:
+            if obj.parent:
+                obj.tree = obj.parent.tree
+            else:
+                obj.tree = Tree.objects.create(quota=request.user.quota)
```
brianhelba, line 21: What if a tree is already selected?

dkc/core/admin/folder.py
```diff
@@ -9,3 +11,15 @@ class FolderAdmin(admin.ModelAdmin):
...
+                obj.tree.grant_permission(
+                    PermissionGrant(user_or_group=request.user, permission=Permission.admin)
```
brianhelba, line 23: What if we want to grant this to another user or not grant it at all?

brianhelba (commented):
I think this is fine as a temporary usability measure, but if we start relying on this, it's an indication that we need to start building some server-rendered staff pages (which can provide a lot more safety and convenience than the admin). We shouldn't abstract away the power of developers to make any change that the models technically permit.

For example here, it should ideally be possible for a developer to create a tree with no permission grants, or granted to a different user. If there's not adequate admin tooling to manage permission grants on Trees, that's something we can improve more directly.

dkc/core/admin/folder.py
zachmullen: What do you mean already selected? Tree is always a computed property, it can't be manually selected even from the admin page.

dkc/core/admin/folder.py
zachmullen: I'm fine to omit this, this is just an attempt to not have a 500 when creating a folder through the admin page.

dkc/core/admin/folder.py
brianhelba: Oh, I see it's `editable=False`. I thought we had a little more discretion to set this up manually in the admin.

dkc/core/admin/folder.py
brianhelba: What will 500? I would expect a 403 somewhere at worst.

Though can't an admin already do anything, even without explicit permission?

zachmullen: On current master, trying to create a new folder via the admin page gives a 500 because it attempts to save a folder with no tree attached. @danlamanna ran into this bug earlier today. We could maybe try to preserve purity by making people manually create a Tree object and then attach a folder to it, but personally I'd rather they always get created together atomically.

brianhelba: Sorry, I just mean is it valid to omit `obj.tree.grant_permission`?

zachmullen: Yes, I believe that's valid.

zachmullen:
I removed the permission grant and changed the quota to use the creator's quota rather than the request user, since the admin page makes administrators select a creator.

dkc/core/admin/folder.py
```diff
@@ -9,3 +10,12 @@ class FolderAdmin(admin.ModelAdmin):
...
 
     search_fields = ['name', 'id']
+
+    @transaction.atomic
+    def save_model(self, request, obj: Folder, form, change: bool):
+        if not change:
+            if obj.parent:
+                obj.tree = obj.parent.tree
+            else:
+                obj.tree = Tree.objects.create(quota=obj.creator.quota)
```
zachmullen, line 20: @brianhelba it might make sense to grant admin permission to `obj.creator`, but I'll let you decide.

brianhelba: I agree this makes more sense.

brianhelba (approved):

Issue #124, opened 2021-01-21
Pre-validation of file uploads in SPA


zachmullen:

zachmullen:
Moving to milestone 2, since we haven't finalized a design of how to best accomplish this. Hoping to do so at our next planning meeting.

mgrauer:
Closed by https://github.com/girder/dkc-next/pull/140.

Issue #139, opened 2021-02-03
Policy enforcement in file and blob creation


zachmullen:
Policies we to enforce around uploading:

- [ ] The size of the blob that is attached to a file matches the declared size of the file.
- [x] File `size` values are immutable, at least via REST.
- [ ] The `blob` field of a file is immutable. It can be changed from no file to having a file, but not from one file to another.
- [ ] Blobs can only be uploaded if they correspond to a pre-initialized file. This will need changes in both S3FF and this repository. See https://github.com/girder/django-s3-file-field/issues/204

mgrauer:
Security focused follow up to #124.

zachmullen:
Third checkbox here is met by #166

PR #140, opened 2021-02-03
Make file creation and validation happen prior to blob upload

zachmullen:
This is the first of two PRs associated with #124 

This provides the upload flow changes so that all uploads are pre-approved. It does not provide any of the security improvements we'd like. I may wait for https://github.com/girder/django-s3-file-field/issues/204 to land before tackling the policy enforcement side of this, as it could simplify the implementation here. #139 represents the work needed on the second PR.


dkc/core/migrations/0009_file_pre_creation.py
```diff
@@ -0,0 +1,28 @@
...
+            model_name='file',
+            name='size',
+            field=models.PositiveBigIntegerField(),
+        ),
+        migrations.AlterField(
+            model_name='quota',
+            name='user',
+            field=models.OneToOneField(
+                on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL
+            ),
```
zachmullen, line 26: @brianhelba this seems to have slipped in from a change on a prior branch that wasn't captured in a migration.

mgrauer:
Fixes #124.

dkc/core/rest/file.py
```diff
@@ -91,4 +93,6 @@ def perform_create(self, serializer: FileSerializer):
     @action(detail=True)
     def download(self, request, pk=None):
         file = get_object_or_404(File, pk=pk)
-        return HttpResponseRedirect(file.blob.url)
+        if file.blob:  # FileFields are falsy when not populated with a file
```
brianhelba, line 96:
```suggestion
        if file.blob:  # FieldFile are falsy when not populated with a file
```

The actual instantiated field value is a `FieldFile`: https://docs.djangoproject.com/en/3.1/ref/models/fields/#filefield-and-fieldfile

dkc/core/rest/file.py
```diff
@@ -91,4 +93,6 @@ def perform_create(self, serializer: FileSerializer):
...
+            return HttpResponseRedirect(file.blob.url)
+        return FileResponse(BytesIO(), filename=file.name, content_type=file.content_type)
```
brianhelba, line 98: If the file hasn't been uploaded yet, I think this should return a 202, 204, or 404 response. To me, a partially uploaded file shouldn't be considered a fully valid resource by end API users.

dkc/core/tests/factories.py
```diff
@@ -54,6 +54,7 @@ class Meta:
     name = factory.Faker('file_name')
     description = factory.Faker('paragraph')
     blob = factory.django.FileField(data=b'fakefilebytes', filename='fake.txt')
+    size = factory.LazyAttribute(lambda obj: obj.blob.size)
```
brianhelba, line 57:
```suggestion
    size = factory.SelfAttribute('blob.size')
```

dkc/core/rest/file.py
zachmullen: Good point, I'm not sure why I didn't think of something like 204. I wanted to specifically avoid 4xx, thinking of someone trying to sync a tree with a partially uploaded file, and I didn't want to break the sync operation if a pending file exists.

dkc/core/tests/test_file_rest.py
```diff
@@ -16,3 +18,50 @@ def test_file_list_default_ordering(admin_api_client, folder, file_factory):
...
+            'name': 'test.txt',
+            'size': 42,
+        },
+    )
+    assert resp.status_code == 201
+    assert resp.data['size'] == 42
+    folder.refresh_from_db()
+    assert folder.size == 42
+    saved_file = File.objects.get(pk=resp.data['id'])
+    assert bool(saved_file.blob) is False
```
brianhelba, line 39:
```suggestion
    assert not saved_file.blob
```

dkc/core/tests/test_file_rest.py
```diff
@@ -16,3 +18,50 @@ def test_file_list_default_ordering(admin_api_client, folder, file_factory):
...
+            'folder': folder.id,
+            'name': 'test.txt',
+            'size': 42,
+        },
+    )
+    assert resp.status_code == 201
+    assert resp.data['size'] == 42
+    folder.refresh_from_db()
+    assert folder.size == 42
+    saved_file = File.objects.get(pk=resp.data['id'])
```
brianhelba, line 38:
```suggestion
    saved_file = File.objects.get(id=resp.data['id'])
```

In principle, the internal primary key field could be different than `id`. Since we're operating from the perspective of an external client here, which only knows about the concrete `id` field, that's what we should use to look up.

This isn't really important here, but just a thought for writing future code.

dkc/core/migrations/0009_file_pre_creation.py
brianhelba: Yeah, I'll add something to test for that in the future.

dkc/core/rest/file.py
brianhelba: I don't think I've fully thought this through, but don't we want file upload to appear to be atomic from the perspective of read-only API endpoints? That is, the file will be non-existent / invisible until the blob (and potentially the hashsum?) is ready.

zachmullen: Perhaps so, but that's a lot more complicated to engineer. If you have a design in mind, please detail it.

dkc/core/tests/test_file_rest.py
```diff
@@ -16,3 +18,50 @@ def test_file_list_default_ordering(admin_api_client, folder, file_factory):
...
+
+@pytest.mark.django_db
+def test_file_rest_set_blob(admin_api_client, pending_file, s3ff_field_value):
+    resp = admin_api_client.patch(
+        f'/api/v2/files/{pending_file.id}', data={'blob': s3ff_field_value}
+    )
+    assert resp.status_code == 200
+    assert resp.data['size'] == pending_file.size
+    pending_file.refresh_from_db()
+    assert bool(pending_file.blob) is True
```
brianhelba, line 67:
```suggestion
    assert pending_file.blob
```

dkc/core/migrations/0009_file_pre_creation.py
brianhelba: https://github.com/girder/cookiecutter-django-girder/pull/125 will ensure this doesn't happen again, once we pull it down.

dkc/core/models/file.py
```diff
@@ -34,7 +34,7 @@ class Meta:
         ],
     )
     description = models.TextField(max_length=3000, blank=True)
-    size = models.PositiveBigIntegerField(editable=False)
+    size = models.PositiveBigIntegerField()
     content_type = models.CharField(max_length=255, default='application/octet-stream')
     blob = S3FileField()
```
brianhelba, line 39:
```suggestion
    blob = S3FileField(blank=True)
```

This is only enforced for forms (including Django admin), but technically this field is now optional.

dkc/core/rest/file.py
brianhelba: As a start, we could filter the `FileViewSet` queryset for listing / details to always exclude anything without a `blob`; that would cause partial files to effectively be invisible in a folder listing. However the folder itself would already have its size incremented, so an attentive observer would notice that certain folder sizes were listed as larger as the sum of their contents.

From a user experience perspective, I'm not sure which is preferable.

I agree that to have fully robust atomicity, we'd have to maintain a separate set of quota reservations or have more complicated logic for reporting effective folder sizes.

dkc/core/rest/file.py
```diff
@@ -64,7 +65,8 @@ def _validate_unique_folder_siblings(self, attrs):
 
 class FileUpdateSerializer(FileSerializer):
     class Meta(FileSerializer.Meta):
-        read_only_fields = FileSerializer.Meta.read_only_fields + ['folder']
+        fields = FileSerializer.Meta.fields + ['blob']
+        read_only_fields = FileSerializer.Meta.read_only_fields + ['folder', 'size']
```
brianhelba, line 69: Right now, there's nothing that prevents subsequent PATCH requests from continuing to overwrite the `blob` again and again. Is this something that will be addressed in the future?

dkc/core/models/file.py
```diff
@@ -84,12 +84,8 @@ def has_permission(self, user: User, permission: Permission) -> bool:
 
 @receiver(models.signals.pre_save, sender=File)
 def _file_pre_save(sender: Type[File], instance: File, **kwargs):
-    # TODO if we allow changing a file's blob, we also need to update the size
-
+    # TODO if we allow changing a file's blob & size, we'll need more logic here
     if not instance.pk:
-        # TODO: Test how this works in minio
-        instance.size = instance.blob.size
```
brianhelba, line 91: We are currently trusting users that the blob's actual size is the same as the declared size.

In the future, we should be able to depend on S3FF to authoritatively guarantee the exact size of the uploaded object (currently, this logic isn't complete in S3FF, but it can be finished) after the upload completes. However, for full functionality here, we still need to link a particular S3FF upload to this particular field *in advance* (which is https://github.com/girder/django-s3-file-field/issues/204).

As a temporary alternative to at least maintain the accuracy of `instance.size`, we could continue to re-set or re-verify the actual blob size `instance.blob.size` (which calls out to S3) vis-a-vis the database size `instance.size` any time the blob is changed.

dkc/core/tests/test_file_rest.py
```diff
@@ -16,3 +18,50 @@ def test_file_list_default_ordering(admin_api_client, folder, file_factory):
...
+    assert folder.size == 42
+    saved_file = File.objects.get(id=resp.data['id'])
+    assert not saved_file.blob
+
+
+@pytest.mark.django_db
+def test_file_rest_download_pending_file(admin_api_client, pending_file):
+    """Test downloading a file prior to its blob being set does something sane."""
+    resp = admin_api_client.get(f'/api/v2/files/{pending_file.id}/download')
+    assert resp.status_code == 200
```
brianhelba, line 46: This test will need to change to a 204.

brianhelba (changes requested):
Finished my review. Should be good to go after the remaining comments are addressed.

dkc/core/rest/file.py
zachmullen: Yes, https://github.com/girder/dkc-next/issues/139 captures this and other policies we want to enforce.

dkc/core/models/file.py
zachmullen: #139 also encompasses this issue.

zachmullen:
@brianhelba this is ready for final review.

brianhelba (approved):

PR #145, opened 2021-02-05
Make quota exceeded failures 400 errors in REST

### LINKED ISSUE #144: Make quota exceeded failures on file creation a 400 rather than 500
Currently this comes out as a model-layer ValidationError, which is treated as 500 in the REST layer.

zachmullen:
Fixes #144

### DIFF
--- a/dkc/core/exceptions.py
+++ b/dkc/core/exceptions.py
@@ -1,0 +1,8 @@
+from django.core.exceptions import ValidationError
+
+
+class QuotaLimitedError(ValidationError):
+    """Raised when a quota would be exceeded by an increment operation."""
+
+    def __init__(self, message='Root folder size quota would be exceeded.', *args, **kwargs):
+        super().__init__(message, *args, **kwargs)
--- a/dkc/core/exceptions.py
+++ b/dkc/core/exceptions.py
@@ -4,5 +4,5 @@ from django.core.exceptions import ValidationError
 class QuotaLimitedError(ValidationError):
     """Raised when a quota would be exceeded by an increment operation."""
 
-    def __init__(self, message='Root folder size quota would be exceeded.', *args, **kwargs):
+    def __init__(self, message='Tree size quota would be exceeded.', *args, **kwargs):
         super().__init__(message, *args, **kwargs)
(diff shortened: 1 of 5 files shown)


dkc/core/rest/file.py
```diff
@@ -86,7 +87,12 @@ def perform_create(self, serializer: FileSerializer):
         user: User = self.request.user
         if not folder.has_permission(user, permission=Permission.write):
             raise PermissionDenied()
-        serializer.save(creator=user)
+        try:
+            serializer.save(creator=user)
+        except QuotaLimitedError:
+            raise serializers.ValidationError(
+                {'size': ['This file would exceed the size quota for this folder.']}
```
zachmullen, line 94: If there's an easier way to achieve this response structure, let me know. I want it to mirror the structure of the other types of validation failure responses.

dkc/core/exceptions.py
```diff
@@ -0,0 +1,8 @@
+from django.core.exceptions import ValidationError
+
+
+class QuotaLimitedError(ValidationError):
+    """Raised when a quota would be exceeded by an increment operation."""
+
+    def __init__(self, message='Root folder size quota would be exceeded.', *args, **kwargs):
```
brianhelba, line 7: Will this message be directly displayed to any users? If not, maybe say:
> Tree size quota would be exceeded

instead, since that's more accurate and helpful to developers.

brianhelba (approved):

dkc/core/rest/file.py
brianhelba: If the error occurs within a serializer field validator, the field will be automatically set. Since it's not thrown during validation, unfortunately it won't.

An alternative implementation could implement an internal API to determine whether a given allocation would exceed the quota. We'd then have to either do the entire API operation atomically, or we'd have to re-check the same thing on save anyway.

Since end users ultimately get the right response this way, and it seems simpler to implement (though slightly less "correct", since we want to try to do all validation in the serializer), I think the current way is best.

brianhelba: Actually, I'll weaken my position: I think the current way is adequate. The alternative would require a new File/Folder/Quota API, but probably won't be any more complex on the DRF layer, since we'd no longer catch exceptions during `.save`.

PR #147, opened 2021-02-05
Add response OpenAPI schema for file download

zachmullen:
This is a very minor addendum to #140 to document the expected HTTP responses for anyone interested in building a client.

### DIFF
--- a/dkc/core/rest/file.py
+++ b/dkc/core/rest/file.py
@@ -2,6 +2,7 @@ from django.contrib.auth.models import User
 from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseRedirect
 from django.shortcuts import get_object_or_404
+from drf_yasg.utils import swagger_auto_schema
 from rest_framework import serializers
 from rest_framework.decorators import action
 from rest_framework.response import Response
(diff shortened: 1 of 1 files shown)


dkc/core/rest/file.py
```diff
@@ -88,9 +89,15 @@ def perform_create(self, serializer: FileSerializer):
             raise PermissionDenied()
         serializer.save(creator=user)
 
-    # TODO figure out how to indicate this response type in the OpenAPI schema
+    @swagger_auto_schema(
+        responses={
+            204: 'This file is pending or has no associated content.',
```
brianhelba, line 94: I'm not sure if "no associated content" is an actual state that we support. I *think* that uploading a 0-byte file will still save an object to S3 and get downloaded via a redirect URL.

brianhelba (approved):

dkc/core/rest/file.py
zachmullen: Girder 3 supported 0 byte files but didn't represent them in the assetstore, they were always special cased. I am not sure whether that's good or bad from an implementation perspective, however I do think supporting them is a good idea since empty files are a valid concept and exist in the wild.

brianhelba: Assuming that we properly support 0-byte files, a 204 will not be the response for one.

So, I'd just suggest rewording this
```suggestion
            204: 'This file is pending and has no associated content.',
```

PR #149, opened 2021-02-10
Add support for coverage

zachmullen:


.coveragerc
```diff
@@ -0,0 +1,4 @@
+[run]
```
brianhelba, line 1: Can you put this in `tox.ini` or `pyproject.toml`: https://coverage.readthedocs.io/en/coverage-5.0/config.html ?

I'd typically prefer `pyproject.toml`, since that's where the tooling config ecosystem is gradually moving toward, but you just need to also install `coverage[toml]`.

In this case, I have no net preference.

brianhelba (approved):

PR #153, opened 2021-02-11
Authorized uploads

zachmullen:
This provides a simple implementation of authorized uploads using a time-bounded digitally signed value.

Things I'd like reviewers to provide opinions on in particular:

1. ~~In Girder 3, authorized uploads allowed only a single file to be uploaded. This implementation allows any number of files to be uploaded to the destination folder. To me, I think this is an improvement, but it represents a policy change so I want to make sure people are OK with that.~~
1. ~~Right now, this does nothing to finalize the authorized upload. I can imagine adding another endpoint that deletes the authorized upload once the user has uploaded all their files, and calling it would also email the user who originally created the authorization. Is it OK to add this in a separate PR, or do we think this is incomplete without that capability? (See next point.)~~
1. ~~Is it appropriate to put the authorized upload creation endpoint under `/folders`? It may make sense to have a separate top-level view set for authorized uploads, especially if / when we add a `DELETE /authorized_uploads/{id}` to close the loop.~~
1. ~~Is the use of the new permission class acceptable? It's theoretically possible to put the entire validation logic inside the new permission class, but there's a few downsides to that, one of which is that we can't provide these nice error messages about exactly why the validation is failing. In the case of expired authorizations in particular, I think it's important for users to know the specific reason they're being 403ed.~~
1. ~~Finally, we could get away without any model to store this, and just sign all the information needed to authorize the upload. However, that would remove the ability to revoke and finalize an authorized upload. Given how little code it takes to have this model in place, I think it's worth it to support those capabilities.~~


dkc/core/rest/file.py
```diff
@@ -83,11 +101,39 @@ def get_serializer_class(self):
             return FileUpdateSerializer
         return FileSerializer
 
+    def _validate_authorized_upload(self, authorization: str, folder: Folder) -> User:
+        try:
+            upload_id = Signer().unsign(authorization)
+        except BadSignature:
+            logger = logging.getLogger('django.security.SuspiciousOperation')
+            logger.exception('Authorized upload signature tampering detected')
```
brianhelba, line 109: Should this be a `logging.warning` instead?

zachmullen: Verdict: yes. It will log stack trace and locals.

zachmullen:
Today's discussion addressed these points to my satisfaction, so this is ready for review.

dkc/core/rest/authorized_upload.py
```diff
@@ -0,0 +1,113 @@
...
+
+class AuthorizedUploadSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = AuthorizedUpload
+        fields = ['id', 'created', 'creator', 'expires', 'folder', 'signature']
+
+    signature: str = serializers.SerializerMethodField()
+
+    def get_signature(self, upload: AuthorizedUpload) -> str:
+        return Signer().sign(str(upload.id))
```
zachmullen, line 32: It occurs to me that there's a slight problem with just signing the ID without scoping it to a specific model type -- if the server ever started signing IDs for other models, someone could pretend that a signature for model foo with ID 123 was a signature for model bar with ID 123 and the server would accept it. So I think I will change this to either use `signing.dumps` and include the model, or else sign a string like e.g. `'AuthorizedUpload.123'`.

zachmullen: The general principle here is that any features making use of the same signing key should have a disjoint set of possible values to be signed.

dkc/core/rest/authorized_upload.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @swagger_auto_schema(
+        responses={201: AuthorizedUploadSerializer},
+        request_body=CreateAuthorizedUploadSerializer,
+    )
+    def create(self, request, pk=None):
+        """Authorize an upload to a folder on your behalf."""
+        serializer = CreateAuthorizedUploadSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        folder = get_object_or_404(Folder, pk=serializer.validated_data['folder'])
```
brianhelba, line 59: I don't think an invalid `folder_id` should trigger a 404 here. I think that would be a 400.

dkc/core/models/authorized_upload.py
```diff
@@ -0,0 +1,34 @@
...
+
+class AuthorizedUpload(models.Model):
+    folder = models.ForeignKey(
+        Folder, on_delete=models.CASCADE, related_name='authorized_uploads', editable=False
+    )
+    creator = models.ForeignKey(
+        User, on_delete=models.CASCADE, related_name='authorized_uploads', editable=False
+    )
+    created = CreationDateTimeField()
+    expires = models.DateTimeField(editable=False)
```
brianhelba, line 17: Could `expires` have a default value function, with the logic (currently in the view layer):
```
return timezone.now() + timedelta(days=settings.DKC_AUTHORIZED_UPLOAD_EXPIRATION_DAYS)
```

dkc/core/rest/authorized_upload.py
```diff
@@ -0,0 +1,106 @@
...
+    )
+    def completion(self, request, pk=None):
+        """Mark an authorized upload as complete and notify its originator."""
+        serializer = CompleteAuthorizedUploadSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        upload = self.get_object()
+
+        try:
+            upload.verify_signature(serializer.validated_data['authorization'])
```
brianhelba, line 86: It might make more sense to put this logic into the `CompleteAuthorizedUploadSerializer`, if it's possible.

dkc/core/rest/authorized_upload.py
```diff
@@ -0,0 +1,106 @@
...
+        except signing.BadSignature:
+            logger = logging.getLogger('django.security.SuspiciousOperation')
+            logger.warning('Authorized upload signature tampering detected.')
+            raise PermissionDenied('Invalid authorization signature.')
```
brianhelba, line 90: A garbage signature is probably a 400 / `ValidationError`, rather than a 403 / `PermissionDenied`. If we put this in the serializer, this should get fixed naturally.

dkc/settings.py
```diff
@@ -41,6 +41,8 @@ class DkcMixin(ConfigMixin):
     ACCOUNT_AUTHENTICATION_METHOD = 'username_email'
 
     DKC_DEFAULT_QUOTA = 3 << 30  # 3 GB
+    DKC_AUTHORIZED_UPLOAD_EXPIRATION_DAYS = 7
+    DKC_SPA_URL = 'https://dkc-next-web.girderops.net/'
```
brianhelba, line 45:
```suggestion
    DKC_SPA_URL = values.Value(environ_required=True)
```

This will require us to set `DJANGO_DKC_SPA_URL` at deployment time, allowing it to be determined by the deployment code.

---

Within `DevelopmentConfiguration`, I'd add:
```
DKC_SPA_URL = values.Value('http://localhost:8080/`)
```
which will give the value a sensible default, but still allow it to be overridden via environment variable.

---

Within `TestingConfiguration`, you can set:
```
DKC_SPA_URL = 'http://dkc-web.test/'
```
which will ensure it's set if it ever gets executed in a test, and gives us a deterministic value if we ever want a test to make assertions about it.

dkc/core/rest/authorized_upload.py
```diff
@@ -0,0 +1,106 @@
...
+    permission_classes = [CanDeleteAuthorization]
+    serializer_class = AuthorizedUploadSerializer
+
+    @swagger_auto_schema(
+        responses={201: AuthorizedUploadSerializer},
+        request_body=CreateAuthorizedUploadSerializer,
+    )
+    def create(self, request, pk=None):
+        """Authorize an upload to a folder on your behalf."""
+        serializer = CreateAuthorizedUploadSerializer(data=request.data)
```
brianhelba, line 57: Could the `AuthorizedUploadSerializer` work just as well here? Are there any other non-read-only fields on it, other than `folder`?

`print(repr(AuthorizedUploadSerializer()))` should be useful to find out.

dkc/core/models/authorized_upload.py
```diff
@@ -0,0 +1,34 @@
...
+        return signing.dumps({'scope': 'authorized_upload', 'id': self.id})
+
+    def verify_signature(self, signature: str) -> None:
+        """Verify the signature associated with this instance.
+
+        Raises `BadSignature` if message tampering occurred, or if the scope
+        of the signed message does not match this feature, or if the signed ID does
+        not match the ID of this instance.
+        """
+        signed_obj = signing.loads(signature)
```
brianhelba, line 30: `signing.dumps` uses a `TimestampSigner`. We could check the `max_age=` (and potentially fail) right here. That would remove some more logic from the view layer.

dkc/core/models/authorized_upload.py
brianhelba: Alternatively, it's not really necessary to store `expires` as a column at all, since we know `created` and `DKC_AUTHORIZED_UPLOAD_EXPIRATION_DAYS`. It could just be a computed `@property`.

dkc/core/rest/file.py
```diff
@@ -83,11 +101,42 @@ def get_serializer_class(self):
...
+            )
+        if upload.folder.id != folder.id:
+            raise PermissionDenied('This upload was authorized to a different folder.')
+
+        return upload.creator
+
     def perform_create(self, serializer: FileSerializer):
         folder: Folder = serializer.validated_data['folder']
-        user: User = self.request.user
+        if 'authorization' in self.request.data:
```
brianhelba, line 133: I think it's unsafe to look at `self.request.data` directly. Given that POST bodies are JSON and auto-decoded, it might not even be a `str` (which is what the parameter type of `_validate_authorized_upload` expects).

We can tack an extra write-only field on `FileSerializer` to obtain this value safely.

brianhelba: Also, I'm hoping that maybe a custom DRF field would allow the `signing.loads` to be abstracted within the serializer, so we're always guaranteed to receive a valid signed value as output.

dkc/core/models/authorized_upload.py
brianhelba: If we do `signing.dumps` in the serializer layer instead, that would negate this suggestion.

dkc/core/rest/file.py
```diff
@@ -83,11 +101,42 @@ def get_serializer_class(self):
             return FileUpdateSerializer
         return FileSerializer
 
+    def _validate_authorized_upload(self, authorization: str, folder: Folder) -> User:
+        try:
+            signed_obj = signing.loads(authorization)
+        except signing.BadSignature:
+            logger = logging.getLogger('django.security.SuspiciousOperation')
+            logger.warning('Authorized upload signature tampering detected.')
+            raise PermissionDenied('Invalid authorization signature.')
```
brianhelba, line 110: Again, probably a 400, not 403.

dkc/core/rest/file.py
```diff
@@ -83,11 +101,42 @@ def get_serializer_class(self):
...
+
+        if signed_obj.get('scope') != 'authorized_upload':
+            raise PermissionDenied('Invalid signature scope.')
```
brianhelba, line 113: This is probably also a 400. It's invalid input; they passed the wrong (validly signed) type of token.

dkc/core/templates/email/authorized_upload_complete.txt
```diff
@@ -0,0 +1,4 @@
+This email is to let you know that a user has completed an upload you authorized
```
brianhelba, line 1:
```suggestion
This email is to let you know that a user has completed an upload which you authorized
```

dkc/core/templates/email/authorized_upload_complete.txt
```diff
@@ -0,0 +1,4 @@
...
+in the folder "{{ folder.abs_path }}".
+
+The files they uploaded can be seen here: {{ folder_url }}
```
brianhelba, line 4:
```suggestion
The files that they uploaded can be seen here: {{ folder_url }}
```

dkc/core/templates/email/authorized_upload_complete.html
```diff
@@ -0,0 +1,21 @@
+<html>
+
+<body style="max-width: 560px; color: #525151; background-color: #f9f5f5; font-family: helvetica; margin: 0 auto;">
+    <center style="background: #fff; border-radius: 5px; box-shadow: 0 4px 20px #ddd; padding: 25px;">
+        <div style="background-color: #eee; color: #474d52; padding: 14px; margin: 0 0 30px;">
+            &check; Authorized upload complete
+        </div>
+        <p>
+            This email is to let you know that a user has completed an upload you authorized in the folder:
```
brianhelba, line 9:
```suggestion
            This email is to let you know that a user has completed an upload that you authorized in the folder:
```

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+        '/api/v2/files',
+        data={
+            'name': 'foo.txt',
+            'size': 123,
+            'folder': authorized_upload.data['folder'],
+            'authorization': f'1{authorized_upload.data["signature"]}',
+        },
+    )
+    assert resp.status_code == 403
+    assert resp.json() == {'detail': 'Invalid authorization signature.'}
```
brianhelba, line 42:
```suggestion
    assert resp.data == {'detail': 'Invalid authorization signature.'}
```

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+        '/api/v2/files',
+        data={
+            'name': 'foo.txt',
+            'size': 123,
+            'folder': folder2.id,
+            'authorization': authorized_upload.data['signature'],
+        },
+    )
+    assert resp.status_code == 403
+    assert resp.json() == {'detail': 'This upload was authorized to a different folder.'}
```
brianhelba, line 58:
```suggestion
    assert resp.data == {'detail': 'This upload was authorized to a different folder.'}
```

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+
+from django.utils import timezone
+import pytest
+
+from dkc.core.models import AuthorizedUpload
+
+
+@pytest.fixture
+def authorized_upload(admin_api_client, folder):
+    return admin_api_client.post('/api/v2/authorized_uploads', data={'folder': folder.id})
```
brianhelba, line 11: I think this should be a Factory and operate at the model layer instead. A fixture is part of the surface under test, why go through the API?

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+        '/api/v2/files',
+        data={
+            'name': 'foo.txt',
+            'size': 123,
+            'folder': folder.id,
+            'authorization': authorized_upload.signature,
+        },
+    )
+    assert resp.status_code == 403
+    assert resp.json() == {
```
brianhelba, line 79:
```suggestion
    assert resp.data == {
```

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+        '/api/v2/files',
+        data={
+            'name': 'foo.txt',
+            'size': 123,
+            'folder': folder.id,
+            'authorization': authorized_upload.data['signature'],
+        },
+    )
+    assert resp.status_code == 403
+    assert resp.json() == {'detail': 'This upload authorization has been revoked.'}
```
brianhelba, line 97:
```suggestion
    assert resp.data == {'detail': 'This upload authorization has been revoked.'}
```

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+    assert resp.status_code == 201
+    assert resp.data['creator'] == authorized_upload.data['creator']
+
+
+@pytest.mark.django_db
+def test_authorized_upload_delete_permission(api_client, user, authorized_upload):
+    api_client.force_authenticate(user)
+    resp = api_client.delete(f'/api/v2/authorized_uploads/{authorized_upload.data["id"]}')
+    assert resp.status_code == 403
+    assert resp.json() == {'detail': 'Only the creator of an authorized upload may delete it.'}
```
brianhelba, line 120:
```suggestion
    assert resp.data == {'detail': 'Only the creator of an authorized upload may delete it.'}
```

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+    resp = admin_api_client.post('/api/v2/authorized_uploads', data={'folder': folder.id})
+    assert resp.status_code == 201
+    assert resp.data['folder'] == folder.id
+    assert resp.data['creator'] == admin_api_client.handler._force_user.id
+    assert 'signature' in resp.data
+
+
+@pytest.mark.django_db
+def test_authorized_upload_requires_write_access(api_client, public_folder, user):
+    api_client.force_authenticate(user)
```
brianhelba, line 25: The cookiecutter has [an `authenticated_api_client` fixture](https://github.com/girder/cookiecutter-django-girder/blob/3e416a831e953ec927760cf8a017fd71df199716/%7B%7B%20cookiecutter.project_slug%20%7D%7D/%7B%7B%20cookiecutter.pkg_name%20%7D%7D/%7B%7B%20cookiecutter.first_app_name%20%7D%7D/tests/conftest.py#L18), which for some reason is missing from DKC-next. It might be helpful to add, for this kind of case.

The `user` should be shared between the fixture and this function.

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+        creator=user,
+        folder=folder,
+        expires=timezone.now() + timedelta(days=1),
+    )
+    resp = api_client.post(
+        f'/api/v2/authorized_uploads/{other_upload.id}/completion',
+        data={'authorization': authorized_upload.data['signature']},
+    )
+    assert resp.status_code == 403
+    assert resp.json() == {'detail': 'Invalid authorization signature.'}
```
brianhelba, line 165:
```suggestion
    assert resp.data == {'detail': 'Invalid authorization signature.'}
```

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+
+@pytest.mark.django_db
+def test_authorized_upload_completion(api_client, authorized_upload, mailoutbox):
+    resp = api_client.post(
+        f'/api/v2/authorized_uploads/{authorized_upload.data["id"]}/completion',
+        data={'authorization': authorized_upload.data['signature']},
+    )
+    assert resp.status_code == 204
+    with pytest.raises(AuthorizedUpload.DoesNotExist):
+        AuthorizedUpload.objects.get(pk=authorized_upload.data['id'])
```
brianhelba, line 176:
```suggestion
    assert not AuthorizedUpload.objects.filter(pk=authorized_upload.data['id']).exists()
```

Seems more straightforward to what we're asserting.

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
+from datetime import timedelta
+
+from django.utils import timezone
+import pytest
+
+from dkc.core.models import AuthorizedUpload
+
```
brianhelba, line 7: @brianhelba @jbeezley We should do a final holistic pass, once the implementation is frozen, to ensure that the important cases are getting covered by these tests. I'm sure the coverage report will be helpful.

dkc/core/tests/test_authorized_uploads.py
```diff
@@ -0,0 +1,179 @@
...
+    assert resp.json() == {'detail': 'This upload was authorized to a different folder.'}
+
+
+@pytest.mark.django_db
+def test_authorized_upload_expired(api_client, folder, user_factory):
+    admin = user_factory(is_superuser=True)
+    authorized_upload = AuthorizedUpload.objects.create(
+        creator=admin,
+        folder=folder,
+        expires=timezone.now() - timedelta(hours=1),
```
brianhelba, line 67: This may become moot if we change the implementation, but I think it's better to initialize `created` in the past, rather than try to mess with `expires`. That should result in a more realistic instance, making this test more robust to future implementation changes.

Of course (especially if we're using a Factory), we need to make sure that this isn't swallowed by `CreationDateTimeField` (which does always cooperate with being manually set).

dkc/core/rest/authorized_upload.py
zachmullen: Interesting, can you clarify when you think to use 404 vs. 400 for invalid resource IDs? Is it just based on whether they're in the URL path or some other part of the request?

dkc/core/tests/test_authorized_uploads.py
zachmullen: I had done this prior to making `signature` a model property, but now that it is, there's no need for this. I'll change it.

dkc/core/tests/test_authorized_uploads.py
zachmullen: Yes, I've been using it in development to check myself. It's very nice.

dkc/core/tests/test_authorized_uploads.py
zachmullen: A side note about these fixtures: it would be great if there was an easier way to grab the user out of `admin_api_client` and `authenticated_api_client`; it's possible now but you have to reach into a private symbol. No big deal as long as we're OK with that practice.

dkc/core/tests/test_authorized_uploads.py
zachmullen: I changed this as a result of making `expires` computed. Unfortunately we can't override the `auto_now_add` behavior of the `created` field at creation time, so I am just updating it afterward for this test case.

dkc/core/rest/authorized_upload.py
zachmullen: This has been resolved by switching the serializer, but I'd still like to hear your thoughts on this.

zachmullen:
@brianhelba I've addressed most of this feedback, but I've not attempted the parts related to moving the signing logic into serializers. I don't have a good sense of your vision for that, maybe you could take a crack at it? Otherwise I'd be happy to do a little pair programming on it.

dkc/core/rest/file.py
```diff
@@ -71,10 +91,21 @@ class Meta(FileSerializer.Meta):
...
 
+class CreateWithAuthorizedUpload(BasePermission):
+    def has_object_permission(self, request, view, obj) -> bool:
+        return False
+
+    def has_permission(self, request: Request, view: View) -> bool:
+        # This doesn't actually do the policy enforcement, it simply allows requests
+        # through if they are authorized upload requests. Enforcement is done in the
+        # view's `perform_create` method.
+        return view.action == 'create' and 'authorization' in request.data
```
zachmullen, line 102: I just ran into a security bug that makes it clear that we need to move all permission checking logic out of `perform_create` and into the serializer. As written, someone can pass any `authorization` value, and other validators will be run before the authorization is cryptographically verified. This is problematic because it can leak things like existence of a file name in a folder that a user doesn't have any access to.

zachmullen: Note that this same issue already existed in the POST /folders endpoint, wherein permission checking is done inside `perform_create`. I'll add a ticket to address that.

brianhelba (approved):

PR #156, opened 2021-02-13
Add hash download REST endpoint

### LINKED ISSUE #7: Add support for hashsum download
Use SHA512.

zachmullen:
Fixes #7 

If you guys approve of this endpoint's behavior, I'll add test cases.

### DIFF
--- a/dkc/core/rest/file.py
+++ b/dkc/core/rest/file.py
@@ -131,9 +131,8 @@ class FileViewSet(ModelViewSet):
         qs = File.objects.filter(sha512=sha512).only('blob').order_by()
         qs = File.filter_by_permission(request.user, Permission.read, qs)
 
-        try:
-            file = qs[:1].get()
-        except File.DoesNotExist:
+        file = qs.first()
+        if not file:
             return Response(status=404)
 
         return HttpResponseRedirect(file.blob.url)
--- a/dkc/core/rest/file.py
+++ b/dkc/core/rest/file.py
@@ -71,6 +71,10 @@ class FileUpdateSerializer(FileSerializer):
         read_only_fields = FileSerializer.Meta.read_only_fields + ['folder', 'size']
 
 
+class HashDownloadSerializer(serializers.Serializer):
+    sha512 = serializers.CharField(min_length=128, max_length=128)
+
+
 class FileViewSet(ModelViewSet):
     queryset = File.objects.all()
 
@@ -109,3 +113,27 @@ class FileViewSet(ModelViewSet):
         if file.blob:  # FieldFiles are falsy when not populated with a file
             return HttpResponseRedirect(file.blob.url)
         return Response(status=204)
+
+    @swagger_auto_schema(
+        query_serializer=HashDownloadSerializer,
+        responses={
+            200: None,  # needed to override default
+            302: 'You will be redirected to download the file contents.',
+            404: 'No file exists with the given hash.',
+        },
+    )
+    @action(detail=False, filterset_fields=[], pagination_class=None)
+    def hash_download(self, request):
+        """Download a file based on the hash of its contents."""
+        serializer = HashDownloadSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        sha512 = serializer.validated_data['sha512'].lower()
+        qs = File.objects.filter(sha512=sha512).only('blob').order_by()
+        qs = File.filter_by_permission(request.user, Permission.read, qs)
+
+        try:
+            file = qs[:1].get()
+        except File.DoesNotExist:
+            return Response(status=404)
+
+        return HttpResponseRedirect(file.blob.url)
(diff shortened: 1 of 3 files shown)


dkc/core/rest/file.py
```diff
@@ -109,3 +113,23 @@ def download(self, request, pk=None):
...
+            404: 'No file exists with the given hash.',
+        },
+    )
+    @action(detail=False, filterset_fields=[], pagination_class=None)
+    def hash_download(self, request):
+        """Download a file based on the hash of its contents."""
+        serializer = HashDownloadSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        sha512 = serializer.validated_data['sha512'].lower()
+        qs = self.get_queryset().filter(sha512=sha512).only('blob').order_by()
```
zachmullen, line 131: I added a little bit of extra tweaking to the queryset to slightly optimize the underlying query. Since hash download seems to be the bulk of the workload taking place on DKC, it seems worth this small amount of extra effort.

mgrauer:
I spoke with @thewtex about this recently, to discuss using the ITK test data as a driving use case, and we can follow up with him soon as he is excited to help.

For now, I'd want to know if (1) it is easy to add other checksums and (2) possible to mount this at an alternate API route. We can figure out what those are exactly when we get more specific requirements.

It looks like the current DKC has `sha512` for [ITK test data](https://data.kitware.com/#item/5ea38c6a9014a6d84e2c38a7), but IIUC Matt asked for providing `md5` also so that they can retire other infrastructure hosting via `md5`.

zachmullen:
> IIUC Matt asked for providing md5

I would strongly encourage migration away from MD5 as it has been broken cryptographically (chosen prefix). I would push back against supporting it in dkc-next.

mgrauer:
Also, can you remind me if we are doing file deduplication? What happens to someone's download request if there are two duplicate files (is this even possible given whatever constraints on File you are using?) and a user has permission for one but not the other?

zachmullen:
If there are two duplicate files with the same content and the requesting user has permission on both of them, one is selected arbitrarily. The only notable effect that has would be if the two files have different names, and the client is relying on the content-disposition value to set the name on their end. (In the case of ExternalData I don't think the client does that, but I'm not sure.)

dkc/core/rest/file.py
```diff
@@ -109,3 +113,27 @@ def download(self, request, pk=None):
...
+        serializer = HashDownloadSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        sha512 = serializer.validated_data['sha512'].lower()
+        qs = File.objects.filter(sha512=sha512).only('blob').order_by()
+        qs = File.filter_by_permission(request.user, Permission.read, qs)
+
+        try:
+            file = qs[:1].get()
+        except File.DoesNotExist:
+            return Response(status=404)
```
brianhelba, line 137:
```suggestion
        file = qs.first()
        if not file:
            return Response(status=404)
```

https://docs.djangoproject.com/en/3.1/ref/models/querysets/#first seems more straightforward.

zachmullen: `first` is exactly the behavior I want. Thanks for the tip -- I was reading lots of Django docs for this case and somehow didn't find it.

zachmullen: It turns out, `qs.first()` has a different query output than `qs[:1].get()`. The difference is, `first` adds an `ORDER BY "core_file"."id" ASC` that is not present in the `.get` version. This is [documented behavior](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.query.QuerySet.first):

> If the QuerySet has no ordering defined, then the queryset is automatically ordered by the primary key.

The implementation is:

```python
def first(self):
        """Return the first object of a query or None if no match is found."""
        for obj in (self if self.ordered else self.order_by('pk'))[:1]:
            return obj
```

I don't think that's very sensible behavior, but there it is. In light of that I'm inclined to go back to the old way, just because I went out of the way to tell the database that no sorting was necessary.

zachmullen: Then again, I suspect that because the PK is the clustering index, it won't incur a real performance penalty in practice, it's just irksome that I can't use `first` the way I want.

dkc/core/rest/file.py
```diff
@@ -109,3 +113,26 @@ def download(self, request, pk=None):
...
+            404: 'No file exists with the given hash.',
+        },
+    )
+    @action(detail=False, filterset_fields=[], pagination_class=None)
+    def hash_download(self, request):
+        """Download a file based on the hash of its contents."""
+        serializer = HashDownloadSerializer(data=request.query_params)
+        serializer.is_valid(raise_exception=True)
+        sha512 = serializer.validated_data['sha512'].lower()
+        qs = File.objects.filter(sha512=sha512).only('blob').order_by()
```
brianhelba, line 131: Maybe add a comment here, indicating that you're over-optimizing the query, since it's so common.

brianhelba (approved):

PR #157, opened 2021-02-13
Add werkzeug in dev to enable runserver_plus

zachmullen:
Pushing this as an RFC. After @danlamanna showed off the `--print-sql` mode of `runserver_plus` I used it locally and it's excellent, and I anticipate wanting it more in the future. Adding it to the dev dependencies lets me just modify the docker compose config to add whatever arguments I want.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ setup(
             'django-debug-toolbar',
             'ipython',
             'tox',
+            'werkzeug',
         ]
     },
 )


zachmullen:
RFC here is mostly re: should this be upstreamed to the cookiecutter.

danlamanna:
Some of my experiences with both:

- runserver_plus is better at handling refreshes, it never gives connection refused. Regular runserver gives connection refused for a brief period every time a code change occurs.
- I /feel/ like runserver is better at "staying up" even when bad code changes occur. I haven't attempted to measure this, but runserver practically never dies for me.
- runserver_plus consistently gives a `too many clients` error for postgres after running it for quite a while, so I think it may not clean up correctly when restarting.

brianhelba:
I think this change is great to move upstream to the cookiecutter. @zachmullen Could you make a PR there too?

I also agree that maybe it's premature to switch the default to `runserver_plus` everywhere. Could you file an upstream bug, which quotes the feedback about this?

brianhelba (approved):

zachmullen:
https://github.com/girder/cookiecutter-django-girder/pull/134

PR #164, opened 2021-02-25
Add preliminary CLI

danlamanna:
This is definitely not in the state that I want it, but I've been holding off pushing it for so long that I wanted to get something out. The name `rvt` is arbitrary.

At the moment there are some dummy credentials hardcoded since I wasn't sure what the approach to take with authentication was. 

`rvt sync dkc://<folder-id> ./some/local/path` (download) should work, but upload is half implemented so far.
`rvt ls dkc://<folder-id>` should work, even with the fancy `--tree` option.

### DIFF
--- a/rvt/rvt/__init__.py
+++ b/rvt/rvt/__init__.py
@@ -25,7 +25,6 @@ import toml
 from rvt.models import RemoteFile, RemoteFolder
 from rvt.transfer import download, upload
 from rvt.types import RemoteOrLocalPath, RemotePath
-from rvt.utils import pager, results
 
 FORMAT = "%(message)s"
 logging.basicConfig(format=FORMAT, datefmt="[%X]", handlers=[RichHandler()])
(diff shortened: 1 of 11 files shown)


zachmullen:
@danlamanna could you put the core changes into a separate PR? It seems like they're small and should be easy to approve independently.

rvt/rvt/__init__.py
```diff
@@ -0,0 +1,367 @@
...
+def _maybe_upload_file(ctx, rfolder: RemoteFolder, lpath: Path):
+    rfile = rfolder.file_by_name(ctx, lpath.name)
+
+    if lpath.exists() and rfile and lpath.stat().st_mtime == rfile.modified.timestamp():
+        logger.debug(f'skipping file {lpath} (same mtime).')
+        ctx.skipped_files.append(rfile)
+        return
+
+    logger.info(f'uploading {lpath}')
+    with open(lpath, 'wb') as stream:
```
jbeezley, line 340:
```suggestion
    with open(lpath, 'rb') as stream:
```

danlamanna:
Future work:
- [ ] figure out the best way to test this
- [ ] add a `cp` command for individual files
- [ ] add a `configure` command for persisting credentials (might make a click extension for this)

danlamanna:
@jbeezley can you verify sync, ls, and ls-tree work the way you would expect? then we can get a basic version of this merged.

brianhelba:
You'll need to rebase this before merging.

PR #165, opened 2021-02-25
Allow filtering folders and files by name

danlamanna:


zachmullen (approved):
LGTM once the unused import is removed

danlamanna:
done (no merge permission).

zachmullen:
@brianhelba can we disable the "Restrict who can push to matching branches" checkbox on master, or was that on intentionally?

PR #166, opened 2021-02-25
Compute file SHA512 on upload

zachmullen:
This now computes SHA-512 on files after their blob is set, i.e. at the closure of the upload flow. This PR attempts to provide at-most-once semantics for setting a blob on a file. Computing the checksum is an idempotent task, but it's important for security reasons that a blob is not able to be set more than once, otherwise it would be possible to forge a checksum via a race condition on the celery tasks. With that in mind, reviewers should convince themselves that the use of `select_for_update` sufficiently enforces this policy.

### DIFF
--- a/dkc/core/rest/file.py
+++ b/dkc/core/rest/file.py
@@ -119,7 +119,12 @@ class FileViewSet(ModelViewSet):
 
     def perform_update(self, serializer: FileUpdateSerializer) -> None:
         with transaction.atomic():
-            # We lock this file row in order to make sure blob can only be set once
+            # We lock this file row in order to make sure `blob` can only be set once.
+            # This ensures that we launch at most one `file_compute_sha512` async jobs.
+            # If we didn't have atomicity on this operation, it would be possible to
+            # create a race condition between async hashing jobs that could cause a
+            # mismatch between the blob and the checksum. Aside from security implications,
+            # violations of the immutable blob policy could also cause data integrity failures.
             file: File = File.objects.select_for_update().get(pk=serializer.instance.pk)
             if file.blob and 'blob' in serializer.validated_data:
                 raise serializers.ValidationError({'blob': ["A file's blob may only be set once."]})
--- a/dkc/core/rest/file.py
+++ b/dkc/core/rest/file.py
@@ -2,6 +2,7 @@ from typing import Dict
 
 from django.contrib.auth.models import User
 from django.core.exceptions import PermissionDenied
+from django.db import transaction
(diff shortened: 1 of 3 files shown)


dkc/core/rest/file.py
```diff
@@ -115,6 +117,18 @@ def perform_create(self, serializer: FileSerializer):
                 {'size': ['This file would exceed the size quota for this folder.']}
             )
 
+    def perform_update(self, serializer: FileUpdateSerializer) -> None:
+        with transaction.atomic():
+            # We lock this file row in order to make sure blob can only be set once
```
mgrauer, line 122: This is the what, can you elaborate to provide the why in the comment, as the intent isn't clear to me from reading this?

mgrauer:
Can you explain a bit more about the checksum forging?

I'd also like you to expand the comment that says that you lock the row to say why you are locking the row, I put a single line review at the relevant spot.

dkc/core/rest/file.py
```diff
@@ -115,6 +117,23 @@ def perform_create(self, serializer: FileSerializer):
                 {'size': ['This file would exceed the size quota for this folder.']}
             )
 
+    def perform_update(self, serializer: FileUpdateSerializer) -> None:
+        with transaction.atomic():
```
brianhelba, line 121: Is it always necessary to open a transaction? Can it only be done if this update contains `'blob'`?

dkc/core/tests/test_file_rest.py
```diff
@@ -54,7 +55,8 @@ def test_file_rest_cannot_update_size(admin_api_client, file):
 
 
 @pytest.mark.django_db
-def test_file_rest_set_blob(admin_api_client, pending_file, s3ff_field_value):
+def test_file_rest_set_blob(admin_api_client, pending_file, s3ff_field_value, mocker):
+    mocker.patch.object(file_compute_sha512, 'delay')
```
brianhelba, line 59:
```suggestion
    compute_sha512_spy = mocker.spy(file_compute_sha512, 'delay')
```

This will allow you to also assert below:
```
compute_sha512_spy.assert_called_once_with(pending_file.pk)
```

brianhelba (approved):
Looks good. Two potential improvement ideas inline.

dkc/core/tests/test_file_rest.py
zachmullen: Doesn't `spy` also call the real method? That's what I'm trying to avoid since we don't want to invoke celery tasks in testing.

PR #167, opened 2021-02-28
Fix root folder uniqueness validation on the API

### LINKED ISSUE #162: Incorrect logic in REST-layer folder name validator
[This `UniqueTogetherValidator`](https://github.com/girder/dkc-next/blob/78ddc26f628b53fd0fb7123270848aadbf5578bb/dkc/core/rest/folder.py#L60-L64) has incorrect logic. Namely, if a root folder exists with a given name, that name cannot be used for any other folder, regardless of whether it's a root or not.

Steps to reproduce:

1. Create  root folder named "foo".
2. Create a non-root folder anywhere else named "foo". 

* Expectation: operation succeeds.
* Actual result: HTTP 400 with error "A root folder with that name already exists."

brianhelba:
Fixes #162

### DIFF
--- a/dkc/core/rest/folder.py
+++ b/dkc/core/rest/folder.py
@@ -56,13 +56,6 @@ class FolderSerializer(serializers.ModelSerializer):
                 fields=['parent', 'name'],
                 message=FormattableDict({'name': 'A folder with that name already exists here.'}),
             ),
-            # This could also be implemented as a UniqueValidator on 'name',
-            # but its easier to not explicitly redefine the whole serializer field
-            serializers.UniqueTogetherValidator(
-                queryset=Folder.objects.filter(parent=None),
-                fields=['name'],
-                message=FormattableDict({'name': 'A root folder with that name already exists.'}),
-            ),
             # folder_max_depth and unique_root_folder_per_tree are internal sanity constraints,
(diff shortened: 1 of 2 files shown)


zachmullen (approved):

dkc/core/rest/folder.py
```diff
@@ -71,10 +64,12 @@ def get_access(self, folder: Folder) -> Dict[str, bool]:
         return folder.tree.get_access(self.context.get('user'))
 
     def validate(self, attrs):
+        self._validate_unique_root_name(attrs)
         self._validate_unique_file_siblings(attrs)
```
zachmullen, line 68: It seems like we could just call `_get_instance_name_and_parent_id` once and pass its return values into these two private helpers, instead of calling it inside each one of them.

brianhelba: I did consider this. I like the abstraction of dispatching to sub-validators that take the same `attrs` as the top-level `.validate`. `_get_instance_name_and_parent_id` is just Python, with no DB access, so it doesn't seem a big deal to call twice.

However, I don't feel strongly about this. Let me know if you think it should be changed.

zachmullen: I don't have strong feelings. We can merge.

PR #170, opened 2021-03-11
The Great Migration

zachmullen:


dkc/core/models/file.py
```diff
@@ -43,6 +43,9 @@ class Meta:
     # Prevent deletion of User if it has Folders referencing it
     creator = models.ForeignKey(User, on_delete=models.PROTECT)
 
+    legacy_file_id = models.CharField(max_length=24, default='')
```
brianhelba, line 46:
```suggestion
    legacy_file_id = models.CharField(max_length=24, empty=True, default='')
```

This is necessary on all the fields, to ensure it can be successfully saved in the admin and other forms.

brianhelba: Why not make this nullable?

zachmullen: [Read the Django docs](https://docs.djangoproject.com/en/3.1/ref/models/fields/#null) ;)

zachmullen: `empty` doesn't appear to be a valid kwarg for that constructor. Do you mean `blank=True`?

brianhelba: > [Read the Django docs](https://docs.djangoproject.com/en/3.1/ref/models/fields/#null) ;)

I knew that. :roll_eyes: I'm not sure my why my brain betrays me sometimes.

> `empty` doesn't appear to be a valid kwarg for that constructor. Do you mean `blank=True`?

Yes, that's what I meant.

PR #172, opened 2021-03-22
Make folder deletion asynchronous

zachmullen:
Recursive folder deletion will often take longer than 30 seconds. If clients wish to be clever, they can handle this 202 as they see fit.

### DIFF
--- a/dkc/core/rest/folder.py
+++ b/dkc/core/rest/folder.py
@@ -22,6 +22,7 @@ from dkc.core.permissions import (
     PermissionFilterBackend,
     PermissionGrant,
 )
+from dkc.core.tasks import delete_folder
 
 from .filtering import ActionSpecificFilterBackend, IntegerOrNullFilter
 from .utils import FormattableDict
@@ -205,6 +206,11 @@ class FolderViewSet(ModelViewSet):
             tree = Tree.objects.create(quota=user.quota)
             tree.grant_permission(PermissionGrant(user_or_group=user, permission=Permission.admin))
         serializer.save(tree=tree, creator=user)
+
+    def destroy(self, request, *args, **kwargs):
+        folder = self.get_object()
+        delete_folder.delay(folder.id)
+        return Response(status=202)
 
     @swagger_auto_schema(
         operation_description='Retrieve the path from the root folder to the requested folder.',
--- a/dkc/core/tasks.py
+++ b/dkc/core/tasks.py
@@ -1,6 +1,6 @@
 from celery import shared_task
 
-from dkc.core.models import File
+from dkc.core.models import File, Folder
 
 
 @shared_task()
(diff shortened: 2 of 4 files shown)


dkc/core/tests/test_folder.py
```diff
@@ -131,3 +132,27 @@ def test_increment_size_negative(folder_factory):
...
+def test_folder_delete_recursive(folder, folder_factory, file_factory):
+    child = folder_factory(parent=folder)
+    file = file_factory(folder=folder)
+    folder.delete()
+
+    with pytest.raises(Folder.DoesNotExist):
+        Folder.objects.get(id=child.id)
+
+    with pytest.raises(File.DoesNotExist):
+        File.objects.get(id=file.id)
```
zachmullen, line 158: The two tests above were moved from the `rest` test set to instead test them at the model layer, to avoid the async call.

dkc/core/tasks.py
```diff
@@ -1,10 +1,15 @@
...
 @shared_task()
 def file_compute_sha512(file_id: int):
     file = File.objects.get(pk=file_id)
     file.compute_sha512()
     file.save()
+
+
+@shared_task()
+def delete_folder(folder_id: int):
+    Folder.objects.get(pk=folder_id).delete()
```
brianhelba, line 15: Strictly speaking, we should consistently use either `id` or `pk`. Here, it doesn't actually matter, since we know those are aliased to the same column.

dkc/core/rest/folder.py
```diff
@@ -379,3 +380,8 @@ def public(self, request, pk=None):
         tree.public = serializer.validated_data['public']
         tree.save()
         return Response(serializer.data, status=200)
+
+    def destroy(self, request, *args, **kwargs):
```
brianhelba, line 384: Could you place this method definition further up, above the `@action` methods? I think that'd be an easier code layout for understanding this `ViewSet`

brianhelba (approved):

dkc/core/tasks.py
zachmullen: Could you clarify what you mean? Or perhaps make a more specific recommendation? It's unclear to me when to use each one.

jbeezley, line 15: Do we need any error handling here if this is queued twice?  Given how this will look on the client side, it doesn't seem like a totally unexpected thing to occur.

zachmullen: If I just say "no", is that acceptable?

jbeezley: I guess.  I'm just imagining the sentry alerts getting raised because of it.

zachmullen: I can see this happening if a user deletes a subfolder, then immediately goes up and deletes its parent folder. Otherwise I wouldn't anticipate much noise.

I'd like to punt on this and address it if and when we hear Sentry complaining.

PR #175, opened 2021-03-25
Update plan to standard-0

mgrauer:


mgrauer:
I do not plan to merge this, it is only so I can see some terraform output for the purposes of documenting the db upgrade path.

PR #176, opened 2021-03-26
Add staff view for trees sorted by file count

zachmullen:
This is completely unstyled as of now since I couldn't get the tailwind stuff to work.


dkc/core/templates/staff/trees_by_file_count.html
```diff
@@ -0,0 +1,29 @@
+<!doctype html>
```
brianhelba, line 1: We should be able to use the base from `django-girder-style`. I'll take a look and verify that.

dkc/core/views.py
```diff
@@ -0,0 +1,23 @@
+from django.contrib.admin.views.decorators import staff_member_required
+from django.db import connection
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import render
+
+TREES_BY_FILE_COUNT_Q = """
+SELECT folder.id, folder.name, sq.c as file_count, pg_size_pretty(folder.size) as size
```
brianhelba, line 7: I do not think it's essential to use raw SQL. Running two queries to render an internal dashboard seems far more reasonable.

zachmullen: Cool, your assignment is to tell me what those two queries are and how to run them 👍

dkc/core/views.py
```diff
@@ -0,0 +1,23 @@
...
+) as sq WHERE folder.tree_id = sq.tree_id AND folder.parent_id IS NULL ORDER BY sq.c DESC
+"""
+
+
+@staff_member_required
+def trees_by_file_count_view(request: HttpRequest) -> HttpResponse:
+    with connection.cursor() as cursor:
+        cursor.execute(TREES_BY_FILE_COUNT_Q)
+        rows = cursor.fetchall()
+    return render(request, 'staff/trees_by_file_count.html', {'rows': rows})
```
brianhelba, line 23:
```suggestion
    return render(request, 'core/trees_by_file_count.html', {'rows': rows})
```

Typically, you namespace the templates by the name of the app: https://docs.djangoproject.com/en/3.1/intro/tutorial03/#write-views-that-actually-do-something

dkc/core/views.py
```diff
@@ -1,23 +1,20 @@
...
 
 @staff_member_required
 def trees_by_file_count_view(request: HttpRequest) -> HttpResponse:
-    with connection.cursor() as cursor:
-        cursor.execute(TREES_BY_FILE_COUNT_Q)
-        rows = cursor.fetchall()
-    return render(request, 'staff/trees_by_file_count.html', {'rows': rows})
+    trees_annotated = (
+        Tree.objects.annotate(
+            size=F('quota__used'),
```
zachmullen, line 13: `quota__used` is not the correct thing to use for the size of the tree, as multiple trees can share the same quota.

dkc/core/templates/core/trees_by_file_count.html
```diff
@@ -0,0 +1,29 @@
...
+          <th>Size</th>
+        </tr>
+      </thead>
+      <tbody>
+        {% for tree in trees %}
+        <tr>
+          <td>{{ tree.id }}</td>
+          <td>{{ tree.name }}</td>
+          <td>{{ tree.num_files }}</td>
+          <td>{{ tree.size }}</td>
```
zachmullen, line 23: Note that with your modified query, we no longer get human-readable sizes directly from the query.

danlamanna: https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#filesizeformat

dkc/core/views.py
brianhelba: Fixed.

dkc/core/templates/core/trees_by_file_count.html
brianhelba: Fixed.

zachmullen:
LGTM. I can't give formal approval since I opened.

brianhelba:
@zachmullen PTAL at the last set of changes, to integrate this with the new staff page templates.

dkc/core/templates/core/staff_tree_size.html
```diff
@@ -0,0 +1,38 @@
...
+  table th {
+    border: 1px solid black;
+}
+</style>
+{% endblock %}
+
+{% block body_content %}
+<h1>Tree Sizes</h1>
+
+<table>
```
zachmullen, line 14: Does Tailwind not provide us some nice table styling we could use?

zachmullen:
LGTM

dkc/core/templates/core/staff_tree_size.html
brianhelba: #181

PR #179, opened 2021-03-29
Optimize the FolderAdmin page

brianhelba:
This makes the number of queries constant, instead of linear.


zachmullen (approved):

dkc/core/admin/folder.py
```diff
@@ -6,6 +6,7 @@
 
 @admin.register(Folder)
 class FolderAdmin(admin.ModelAdmin):
+    list_select_related = ['parent']
```
zachmullen, line 9: There's a magical string property for everything

PR #182, opened 2021-03-30
Allow sorting the staff tree list page by size

brianhelba:


dkc/core/views.py
```diff
@@ -12,14 +12,22 @@ def staff_home(request: HttpRequest) -> HttpResponse:
 
 
 @staff_member_required
-def staff_tree_size(request: HttpRequest) -> HttpResponse:
+def staff_tree_list(request: HttpRequest, sort_by) -> HttpResponse:
+    try:
```
brianhelba, line 16: @danlamanna Should I use query params here? This way is easier to build URLs with, it's just slightly annoying that I can't easily specify a default (without creating a separate URL redirect).

dkc/core/templates/core/staff_tree_list.html
```diff
@@ -7,6 +7,22 @@ <h1 class="text-2xl font-bold text-gray-900">
       Tree Sizes
     </h1>
 
+    <div>
```
brianhelba, line 10: I have no ability to style this. I tried and failed, sorry.

dkc/core/views.py
zachmullen: This definitely feels more like it should be a query param. (I could see us adding more parameters in the future.)

brianhelba: Fixed.

zachmullen (approved):

PR #184, opened 2021-04-02
Optimize anonymous user tree access

zachmullen:
This brought the time of listing 7 folders from ~3 sec to around ~250ms in the anonymous user case. This was a simple matter of not correctly detecting anonymous users, for which we have an optimized path.

Let's merge this and I'll follow up with other optimizations in other PRs.

### DIFF
--- a/dkc/core/models/tree.py
+++ b/dkc/core/models/tree.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Dict, List, Optional, Set, Type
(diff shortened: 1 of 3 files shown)


zachmullen:
@jbeezley would it make more sense to change the accepted type from `Optional[User]` to just `User`? Then we could force there to be a 'user' field in the serializer context.

brianhelba:
> would it make more sense to change the accepted type from `Optional[User]` to just `User`?

Good idea. If `user` isn't in the context, let it die immediately at runtime.

brianhelba (approved):

PR #186, opened 2021-04-09
Redirect to the SPA from /

brianhelba:


dkc/urls.py
```diff
@@ -22,6 +23,7 @@
 )
 
 urlpatterns = [
+    path('', RedirectView.as_view(url=settings.DKC_SPA_URL, permanent=True), name='home'),
```
brianhelba, line 26: I think this should be a permanent redirect, but maybe not?

zachmullen (approved):

